PR #36, opened 2019-07-03
Issue #39: Fix memory leak on large bundles

### LINKED ISSUE #39: Investigate and fix memory leak causing validate crash on large bundles

galenatjpl:
To address the memory leak, I did the following (NOT all apply to
memory leak)
* I was able to upgrade to the latest saxon (9.9.1-4).
  There are several method/object changes/incompatibilities with the new
  version…  But it seemed to helped.  While profiling in jvisualVM,
  the saxon threads were always showing up near the top of the list,
  in terms of memory deltas.
* I also refactored some code that closes off the JSON writer objects as
  it goes.  While I didn’t have time to test this out with and without
  this change, it seemed like a possible problem..
* I upgraded to the latest GSON libraries (2.8.5)

Resolves #39


jordanpadams:
@mcayanan do you think you may have a chance early next week to test this? would like to be able to merge this into this sprint. if not, we can wait until @galenatjpl gets back.

mcayanan:
@jordanpadams Currently running a test where I'm outputting the default report style. Just want to make sure that the memory leak isn't showing on there as well before calling this good.

mcayanan:
@jordanpadams @galenatjpl Running where it produces the standard human readable report looked good. However, when trying to produce a json style report, the tool seems to hang. Try running it against the example bundle in:

https://starbase.jpl.nasa.gov/pds4/1900/dph_examples_v1900.zip

When running the tool like so,
`$ ./validate ~V1900/dph_example_archive -Rpds4.bundle -s json -r report.txt`

it seemed to be hanging on my end. Unfortunately I don't have more time to look into this more. I'm particularly interested as to why the JSON report needs to refresh. Seems like flushing the report would be sufficient enough to dump the contents from memory, no?

galenatjpl:
@mcayanan @jordanpadams 
I verified the hang the @mcayanan reported.  However, this hang occurs on both the "master" branch and the "29" branch.  So I think this hang has something to do with the data itself. 
I see this in the output report:
```{
      "status": "FAIL",
      "label": "file:/Users/ghollins/dev/pds/testing/validate/V1900/dph_example_archive/bundle_izenberg_pdart14_meap.xml",
      "messages": [
        {
          "severity": "ERROR",
          "type": "error.label.schematron",
          "line": 11,
          "column": 36,
          "message": "The attribute pds:information_model_version must be equal to the value '1.12.0.0'."
        }
      ],
      "fragments": [],
      "dataContents": []
    }
```

data being validated is at this:

```<Product_Bundle xmlns="http://pds.nasa.gov/pds4/pds/v1"
 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="http://pds.nasa.gov/pds4/pds/v1 https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1700.xsd">
    <Identification_Area>
        <logical_identifier>urn:nasa:pds:izenberg_pdart14_meap</logical_identifier>
        <version_id>1.0</version_id>
        <title>Izenberg PDART 2014 MESSENGER Advanced Products Bundle</title>
        <information_model_version>1.7.0.0</information_model_version>
```

I'm looking into the proper way to work this...

mcayanan:
@galenatjpl @jordanpadams That's interesting. I know that this test bundle ran successfully in previous releases of the tool. So something must have changed in the interim.

mcayanan:
What code base was used in the latest release of validate tool? 

If validate still exists in the pds4-java tree on the JPL github side (at least on master), might be worthwhile to build a version under that tree to see if the same behavior exists.

galenatjpl:
UPDATE:  @jordanpadams pointed out that I should be using the `-f` flag.  After using that, it runs, but still hangs on both `master` and `29`

galenatjpl:
It turns out that the dataset you linked to stalls on all versions of validate I've tried (1.15, master, 29).
This set works however:
https://starbase.jpl.nasa.gov/pds4/1700/dph_example_archive_VG2PLS.zip

So I think it's something about the 1900 dataset that causes it to hang.

However, the 1700 dataset mentioned above does not seem to have enough data to simulate the memory leak. The InSight dataset does however...

galenatjpl:
@mcayanan Per your question/comment, It probably worked pre-1.15, since it doesn't work on 1.15

mcayanan (approved):

mcayanan:
@galenatjpl Okay, in that case, the change is good then. Just wanted to make sure nothing broke that previously worked as a result of this change.

PR #49, opened 2019-07-13
Issue #33: Ignore resources with null systemId

### LINKED ISSUE #33: Validation succeeds despite throwing exception
The validator (1.14.0) sometimes reports no errors even though it has thrown a NullPointerException during the validation process. This leaves me wondering whether there is actually an error in the XML. The exception message doesn't give a line number, so I don't know where in the XML the problem occurs.

Please see the attached output.

[ValidationLog.txt](https://github.com/NASA-PDS-Incubator/validate/files/3346023/ValidationLog.txt)

jordanpadams:
Similar to the functionality in CachedLSResourceResolver.resolveResource,
if systemId is null, we should just return null before trying to resolve it.

Resolves #33


src/main/java/gov/nasa/pds/tools/label/XMLCatalogResolver.java
```diff
@@ -349,6 +349,21 @@ public LSInput resolveResource(String type, String namespaceURI,
         // Ignore. Should not happen!!!!
       }
       
+	  // Similar to CachedLSResourceResolver, if systemId == null
+	  // we won't be able to resolve this resource
+      if (getProblemHandler() != null) {
```
mcayanan, line 354: @jordanpadams Sorry for the confusion. This if block should go below within the:

````
if (systemId == null) {
  return null;
}
````
at line 364 just before returning null

jordanpadams: doh! 👎 for me. this is why I shouldn't be coding anymore!

PR #53, opened 2019-07-17
Issue #24: Retrieve and use latest context products from EN Registry

### LINKED ISSUE #24: Update context check to retrieve and use latest context products from EN Registry
The context reference validation feature that has been added to Validate Tool is just a short term solution. I think in the long run, we will want to go out to the EN Registry to find the latest context products.

Basically there would be an additional class / method added to the Validate source to perform this remote check of the EN Registry.

I would think it should replace the JSON file that comes with the delivered software (backup the old version), that way if they do not have an internet connection, it will validate with the latest based upon when the user last ran the tool.

Requirements: Add a `-u`, update,  flag that forces the tool to try to download the latest set of context products (and replace the existing file).

See also closed pull request on private repo: https://github.jpl.nasa.gov/PDSEN/pds4-java/pull/114

jordanpadams:
Added a flag (-u) that specifies that the latest JSON context should be
download (and replace the existing file).

Modified classes in tools package to make sure all classes should be
inside tools package.

Resolves #24

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,6 +115,14 @@ POSSIBILITY OF SUCH DAMAGE.
         <directory>src/main/resources</directory>
         <filtering>true</filtering>
       </resource>
+      <resource>
+		<filtering>true</filtering>
+		<targetPath>gov/nasa/pds/validate/util</targetPath>
+		<directory>src/main/resources</directory>
+		<includes>
+			<include>validate.properties</include>
+		</includes>
+	  </resource>
     </resources>
 
     <plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -353,6 +353,11 @@ POSSIBILITY OF SUCH DAMAGE.
       <artifactId>log4j</artifactId>
       <version>1.2.17</version>
     </dependency>
+	<dependency>
+      <groupId>org.apache.solr</groupId>
+	  <artifactId>solr-solrj</artifactId>
+	  <version>7.7.2</version>
+	</dependency>
     <dependency>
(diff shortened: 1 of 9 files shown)


jordanpadams:
@danyu-pds since there are no integration/unit tests, can you please provide the data (URL or attached) you used to test this and a test case that can be run?

danyu-pds:
> @danyu-pds since there are no integration/unit tests, can you please provide the data (URL or attached) you used to test this and a test case that can be run?

- Run the script:
bin/validate -u

- Then check the files in resources:
resources/registered_context_products.json.backup
resources/registered_context_products.json

- Also compare registered_context_products.json with the the file which get from the URL:
first use this
https://pds.nasa.gov/services/search/archive-filter?q=product_class:Product_Context&wt=json
to get the number of found:
response |  
-- | --
numFound | 1881

Then use this to get all records:
https://pds.nasa.gov/services/search/archive-filter?q=product_class:Product_Context&wt=json&start=0&rows=1881

danyu-pds:
@jordanpadams 
I have added the SolrJ dependency in the pom.xml. Not sure why you add it again?

jordanpadams:
> @jordanpadams
> I have added the SolrJ dependency in the pom.xml. Not sure why you add it again?

@danyu-pds that was an oversight on my part in my review of the pom.xml. should be good to go now.

danyu-pds (approved):
Check out the branch, modified the pom.xml file and test it. It looks good to me.

Issue #87, opened 2019-09-05
Update validate to apply catalog file and local schemas to discipline schema resolution


jordanpadams:
Per the second portion of integration test `testGithub71`, we use a catalog file to resolve schema locations.

However, it fails if not longer have an internet connection. It appears that it sufficiently resolve the catalog information for the immediate schemas referenced from the label, but does not resolve the catalog information for nested references to schemas, e.g.:

```
  <xsd:import namespace="http://pds.nasa.gov/pds4/pds/v1"
        schemaLocation="https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1700.xsd">
```

For example when I try running validate with the attached data (and modified catalog file) I get this:
```
$ validate-1.16.0-SNAPSHOT/bin/validate -R pds4.label -t ELE_MOM.xml -C catalog.xml
...
Summary:

  2278 error(s)
  2 warning(s)

  Message Types:
    2278         error.table.field_value_data_type_mismatch
    2            warning.label.context_ref_mismatch
```
But when I disconnect from the internet, I get this:
```
Product Level Validation Results

  FAIL: file:/Users/jpadams/Documents/proj/pds/pdsen/workspace/validate/src/test/resources/github71/ELE_MOM.xml
    Begin Schema: https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1700.xsd
      FATAL_ERROR  [error.label.unresolvable_resource]   pds.nasa.gov
    End Schema: https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1700.xsd
    Begin Schema: file:/Users/jpadams/Documents/proj/pds/pdsen/workspace/validate/src/test/resources/github71/test/v1/PDS4_TEST_1700.xsd
      WARNING  [warning.label.schema]   line 24, 127: schema_reference.4: Failed to read schema document 'https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1700.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.
      ERROR  [error.label.schema]   line 39, 62: src-resolve: Cannot resolve the name 'pds:ASCII_Short_String_Collapsed' to a(n) 'type definition' component.
    End Schema: file:/Users/jpadams/Documents/proj/pds/pdsen/workspace/validate/src/test/resources/github71/test/v1/PDS4_TEST_1700.xsd

Summary:

  2 error(s)
  1 warning(s)

  Message Types:
[...]

Issue #84, opened 2019-09-03
Update Validate to allow specification of schema/schematrons in config file


galenatjpl:
/home/xxx/archive_pipeline/tools/validate-1.15.3/bin/validate  --report-style json -c /home/xxx/pipeline/scripts/validate_config.txt  /m/bundle_workspace/insight_cameras -D -R pds4.bundle  2>&1 |
Cannot specify user schemas, schematrons, and/or catalog files with the 'force' flag option

In this case, the user didn't specify '-f', but saw the above error.

This may also be related to #87 .

galenatjpl:
Galen to LFT dataset to Hyun.

galenatjpl:
@hhlee445 I sent the dataset to you over LFT, along with a usage example...

jordanpadams:
@galenatjpl the dataset shouldn't really matter. this error will appear when you try to run bundle validation against any bundle. the config file is what really matters to find out what they are specifying and where we are missing `setForce(false);`, I think we need this in `ValidateLauncher.query(config)`:
```
            if (config.containsKey(ConfigKey.SCHEMA)) {
                // Removes quotes surrounding each pattern being specified
                List<String> list = config.getList(ConfigKey.SCHEMA);
                list = Utility.removeQuotes(list);
                setSchemas(list);
                setForce(false); <<<<<<<<<<<<<<<<<
            }
            if (config.containsKey(ConfigKey.SCHEMATRON)) {
                // Removes quotes surrounding each pattern being specified
                List<String> list = config.getList(ConfigKey.SCHEMATRON);
                list = Utility.removeQuotes(list);
                setSchematrons(list);
                setForce(false);   <<<<<<<<<<<<<<<<<
            }
```

@hhlee445 please try this out and create an integration test with an input config file that tests this. you may just be able to use github71 data

hhlee445:
@jordanpadams just setting setForce(false) still checked the schema/schematron validation, I had to add a separate method to disable schema check in LabelValidator and LocationValidator classes.

jordanpadams:
@hhlee445 has this code been committed?

hhlee445:
I will commit today.





-------- Original message --------

jordanpadams:
@hhlee445 any word on this update? would like to get this in this build if we could.

hhlee445:
Jordan,

I already delivered the code at Sep 19, 2019. I think I forgot to create a PR.



Date: Thursday, September 19, 2019 at 2:29 PM
[...]

PR #113, opened 2019-09-26
Issue #87: Deactivate using online schemas validation (to be closed)

jordanpadams:


jordanpadams:
@hhlee445 I rebased your issue_84 branch to get it inline with the other code commits. that being said, I think the functionality you added was an attempt at fixing #87, not #84.

#84 is really just a bug where the user wants to be able to specify schema/schematron via a config file. #87 is somewhat related to this by because you need to be able to specify schema/schematron in a config, but #84 now needs to apply those schema/schematrons to the actual schema/schematron URL resolver. I am going to update that ticket with some more info a developer from GEO was able to pull out.

PR #157, opened 2019-11-27
issue_156:  Performance Log Metrics and more

### LINKED ISSUE #156: Introduce label and array timing metrics logging
This change adds performance timing logging of array content
and label validation.  The logging will be enabled at verbosity
level 1 (INFO) or below (e.g. DEBUG).

Also, the hooks for experimentation of multi-threading has
been setup at the LabelInFolderRule level.  In other words a
thread pool exists, that can run targets within a folder
concurrently.  However, due to thread safety issues with the
current code and libraries, the thread level is currently set
to one (no multi-threading).  In the future, this can be
experimented with.  A few synchronized locks have been placed
throughout the code, where multi-threading proved problematic
during experimentation.  More may be needed in the future.
When the thread level is set to greater than one, the
performance does improve, but there are failures sometimes,
due to thread saftety issues.  If this can be solved, real
performance gains can be realized.

Some basic GNU parallel scripts have been added in the
src/test directory, as well, for future experimentation into
running validate concurrently at the JVM level.

Also, some library versions have been upgraded in the pom.xml.

galenatjpl:
This change adds performance timing logging of array content
and label validation.  The logging will be enabled at verbosity
level 1 (INFO) or below (e.g. DEBUG).

Also, the hooks for experimentation of multi-threading has
been setup at the LabelInFolderRule level.  In other words a
thread pool exists, that can run targets within a folder
concurrently.  However, due to thread safety issues with the
current code and libraries, the thread level is currently set
to one (no multi-threading).  In the future, this can be
experimented with.  A few synchronized locks have been placed
throughout the code, where multi-threading proved problematic
during experimentation.  More may be needed in the future.
When the thread level is set to greater than one, the
performance does improve, but there are failures sometimes,
due to thread saftety issues.  If this can be solved, real
performance gains can be realized.

Some basic GNU parallel scripts have been added in the
src/test directory, as well, for future experimentation into
running validate concurrently at the JVM level.

Also, some library versions have been upgraded in the pom.xml.

Resolves #156

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,6 +164,7 @@ POSSIBILITY OF SUCH DAMAGE.
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>buildnumber-maven-plugin</artifactId>
+        <version>1.4</version>
         <executions>
           <execution>
             <phase>validate</phase>
@@ -183,8 +184,8 @@ POSSIBILITY OF SUCH DAMAGE.
         </configuration>
       </plugin>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
+        <version>3.1.1</version>
         <executions>
           <execution>
             <id>bin-release</id>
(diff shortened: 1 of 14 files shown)


galenatjpl:
Some interesting performance observations (not conclusive).

TIMINGS:

NEW CODE WIFI LOCATION #1
1m37.385s
1m50.389s

NEW CODE WIFI LOCATION #2 (faster internet than #1)
1m35.443s
1m34.912s
1m34.004s

OLD CODE WIFI LOCATION #1
2m14.229s
2m40.070s

OLD CODE WIFI LOCATION #2 (faster internet than #1)
1m37.496s
1m32.752s

CONCLUSION:
The newer code seems to be faster when the internet speeds are slower.  This may be because certain parts of the code remove reliance on unnecessary URL resolution....

PR #159, opened 2019-12-09
issue_56: Added error message when the definition of two fields in a …

### LINKED ISSUE #56: Validate that Table_Character/Table_Binary fields match the field length definitions in the label
**Is your feature request related to a problem? Please describe.**
Currently, validate does not report when the definition of two fields in a Table_Character overlap. This is a relatively common error by data producers, and should be easy to flag.

**Describe the solution you'd like**
validate should report an error if fields overlap

hhlee445:
…Table_Character overlap.

Resolves #56


jordanpadams:
@hhlee445 can we be sure to add `Resolves #56` in the comment for the PR? that way when it is merged, it automatically closes the ticket for us.

hhlee445:
Resolve #56

PR #174, opened 2020-01-22
Fix bug introduced from #149 fix

### LINKED ISSUE #173: Bug with #149 fix: Product counter is invalid and does not reset after each collection
**Describe the bug**
Product-Level Validation:
1.18.0 and 1.16.0-Snapshot produced the same results, with the exception of 1.18.0 not scanning the labels of the data files themselves (i.e., kgrs_calibrated_spectra_per<N>.xml where N = 1, 2, 3).
 
Bundle-Level Validation:
1.16.0-Snapshot achieved “Pass” across the board. 1.18.0 produced errors/warnings of two types:
1) ERROR  [error.table.records_mismatch]   Number of records read is not equal to the defined number of records in the collection (expected <X>, got <Y>).
2) WARNING  [warning.file.not_referenced_in_label]   File is not referenced by any label
 
Note, that visual inspection of the relevant files pertaining to bundle-level error type 1) show the number of records do match. Furthermore, the files noted in 2) are referenced within the appropriate label.

**To Reproduce**
Run validate against any bundle with more than 1 collection

jordanpadams:
Row counter needs to be reset after each collection is read

Commit also include regression test.

Resolves #173

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/CollectionReferentialIntegrityRule.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/CollectionReferentialIntegrityRule.java
@@ -59,8 +59,6 @@ public class CollectionReferentialIntegrityRule extends AbstractValidationRule {
   
   private String lid = null;
   
-  private int numOfCollections = 0;
-  
   @Override
   public boolean isApplicable(String location) {
     if (Utility.isDir(location)) {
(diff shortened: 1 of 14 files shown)


rchenatjpl:
By "reset counter", you mean something internal, right? I.e. when running validate, NNN should keep incrementing when changing to a different collection:
  PASS: file:/Users/rchen/Desktop/test/bundle_geo_ra/context/mission.phoenix_1.1.xml
        6 product validation(s) completed
  PASS: file:/Users/rchen/Desktop/test/bundle_geo_ra/data_derived/sol101a.xml
        7 product validation(s) completed
That makes good sense to me, and that's what I'm seeing.

jordanpadams:
@rchenatjpl correct. the problem was it was incrementing the counter even when it switch over to "Product Bundle Validation Results:", which doesn't make any sense, and would basically mean you count every product twice.

PR #184, opened 2020-02-03
Implement validate-bundle script

### LINKED ISSUE #128: Develop validate-bundle script for Validation performance improvements using parallelization approach

galenatjpl:
This adds a new helper script, validate-parallel, which allows
for running validation in parallel.
The necessary, related documentation on how to use
validate-parallel is also included in this commit.

Resolves #128


src/main/resources/bin/validate-parallel.sh
```diff
@@ -0,0 +1,130 @@
...
+  echo "         Please ensure this is on your environment PATH, before running this script."
+  echo "         See also:  https://www.gnu.org/software/parallel"
+  echo
+  exit 1
+fi
+
+NUM_CORES=`getconf _NPROCESSORS_ONLN`
+if [[ "${NUM_GROUPS}" -eq "" ]]; then
+  echo "  Splitting up into groups, based on number of cores this machine has..."
+  NUM_GROUPS=`echo $NUM_CORES`
```
jordanpadams: @galenatjpl after testing, does it make sense to use all cores? or was it more efficient to use ~1/2 of the cores available?

galenatjpl: I updated the code to be half-cores by default, and committed this change (along with doc updates).

PR #191, opened 2020-03-07
Add validated product counter to pass/fail reporting and end summary

### LINKED ISSUE #132: Add validated product counter to pass/fail reporting and end summary
Per conversations with GEO and RMS, a counter should be added to pass/fail report dialog to include a counter (validated/to go) in order to be able to monitor status. In addition, in the final output summary, the old VTool included a "validated" number at the end report to show successful validated products

hhlee445:
Added a section in the end summary as

  Number of Passed product(s)/folder(s):  1
  Number of Failed product(s)/folder(s):  4
  Number of Skipped product(s)/folder(s): 0
  Total Number of product(s)/folder(s):   5

A counter for monitoring was added, however, total number of products to go was not included since it takes too much times to implement. 

Resolves #132

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/RuleContext.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/RuleContext.java
@@ -1,16 +1,32 @@
-// Copyright 2006-2018, by the California Institute of Technology.
-// ALL RIGHTS RESERVED. United States Government Sponsorship acknowledged.
-// Any commercial use must be negotiated with the Office of Technology Transfer
-// at the California Institute of Technology.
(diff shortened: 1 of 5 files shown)


jordanpadams:
@hhlee445 is there a reason we include folder in the output message? when i tested it doesn't look like it counts folders?

jordanpadams:
@hhlee445 nevermind. i see now that it adds to the counter for all rule validation. any way we can limit this to just the "Product Level Validation Results" ?

hhlee445:
It will be doable, but it will take a while. The count was incremented for all types of validation.

jordanpadams (approved):

PR #207, opened 2020-04-03
Validate incorrectly fails for ASCII_Integer value in Table_Character

### LINKED ISSUE #206: Validate incorrectly fails ASCII_Integer field that is space-padded (empty) in Table_Character
See https://github.com/NASA-PDS-Incubator/validate/issues/9#issuecomment-607975642

For resolution, we should be sure to include a regression test for both DSV and character tables to make sure both are being checked for the foreseeable future.

hhlee445:
Resolves #206.


src/test/java/gov/nasa/pds/validate/ValidationIntegrationTests.java
```diff
@@ -480,23 +480,6 @@ void testGithub09() {
...
-            this.launcher = new ValidateLauncher();
-            this.launcher.processMain(args4);
-
-            gson = new Gson();
-            reportJson = gson.fromJson(new FileReader(report), JsonObject.class);
-
-            count = this.getMessageCount(reportJson, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey());
-
-            assertEquals(count, 1, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey() + " info/error messages expected.");
-
```
jordanpadams: @hhlee445 this integration test is actually valid and indicates the solution will unfortunately not work. Note in the first row of data for [the file](https://raw.githubusercontent.com/NASA-PDS-Incubator/validate/master/src/test/resources/github09/csv_empty_field_test_INVALID.csv), one of the columns is supposed to be an ASCII_Integer, but is just a space. That is actually invalid and was actually what #9 was all about. You can have spaces surround and integer, but it can't be only whitespace.

```
13, Plains,First_Soil,Soil,First_Soil1_Final,,0,0,13,, ,3,9,"0.406622, -0.859749, 0.275765","0.095748, -0.031037, -0.994659",,,FirstSoil1Final ,
14, Plains,First_Soil,Soil,First_Soil1_Final,,0,0,,14,14,3,9,"0.406622, -0.859749, 0.275765","0.095748, -0.031037, -0.994659",Gusev_Soil,FirstSoil1Final,,

...13,, ,3...
-------^-----
...14,14,3,9...
---------^-----
```

src/test/java/gov/nasa/pds/validate/ValidationIntegrationTests.java
```diff
@@ -695,7 +678,7 @@ void testGithub137() {
 
             count = this.getMessageCount(reportJson, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey());
 
-            assertEquals(count, 2, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey() + " info/error messages not expected.");
+            assertEquals(count, 0, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey() + " info/error messages not expected.");
```
jordanpadams: this failure is valid

src/test/java/gov/nasa/pds/validate/ValidationIntegrationTests.java
```diff
@@ -760,6 +743,57 @@ void testGithub173() {
...
+                    };
+            this.launcher = new ValidateLauncher();
+            this.launcher.processMain(args2);
+
+            gson = new Gson();
+            reportJson = gson.fromJson(new FileReader(report), JsonObject.class);
+
+            count = this.getMessageCount(reportJson, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey());
+
+            assertEquals(count, 0, ProblemType.FIELD_VALUE_DATA_TYPE_MISMATCH.getKey() + " info/error messages not expected.");
```
jordanpadams: This test should actually throw an error because this product is described as a Table_Delimited, but the a couple of values contain all spaces for ASCII_Integer. That is invalid. You can only have all spaces for Table_Character. For Table_Delimited, you can wrap an integer with spaces, they just can be all spaces.

PR #225, opened 2020-05-08
Fix for incorrect validation of number of records (#220)

### LINKED ISSUE #220: Incorrect validation of number of records
**Describe the bug**
v1.22.3 now fails products which are, I believe, completely valid. It seems this is related to the modified check on the number of records in #149.

**To Reproduce**
Create a product with multiple tables, headers, arrays. e.g.

```
        <Header>
            <name>FITS HEADER</name>
            ...
        </Header>
        <Header>
            <name>FITS EXTENSION HEADER</name>
            ...
        </Header>
        <Table_Binary>
            <name>METADATA</name>
            <records>24</records>
            ....
        </Table_Binary>
        <Header>
            <name>FITS EXTENSION HEADER</name>
            ....
        </Header>
        <Array_2D>
            <name>CHANNEL_A_RAW_SCIENCE_DATA</name>
            ....
        </Array_2D>
        <Header>
            <name>FITS EXTENSION HEADER</name>
            ....
        </Header>
        <Array_2D>
            <name>CHANNEL_B_RAW_SCIENCE_DATA</name>
           ....
        </Array_2D>
    </File_Area_Observational>
```
In this case validate gives me an error for the table:
```
ERROR  [error.table.records_mismatch]   table 1: Number of records read is not equal to the defined number of records in the label (expected 24, got 212).
```

hhlee445:
Resolve #220


rchenatjpl:
For file val217/binobs3.xml I think this validated correctly under the previous validate, but now...
% validate -t binobs3.xml
PDS Validate Tool Report
Configuration:
   Version                       1.23.0
   Date                          2020-05-09T23:43:52Z
Parameters:
   Targets                       [file:/Users/rchen/Desktop/binobs3.xml]
   Severity Level                WARNING
   Recurse Directories           true
   File Filters Used             [*.xml, *.XML]
   Data Content Validation       on
   Product Level Validation      on
   Max Errors                    100000
   Registered Contexts File      /PDS4tools/validate/resources/registered_context_products.json
Product Level Validation Results
  FAIL: file:/Users/rchen/Desktop/binobs3.xml
    Begin Content Validation: file:/Users/rchen/Desktop/bin.dat
      ERROR  [error.table.records_mismatch]   table 1: Number of records read is not equal to the defined number of records in the label (expected 3, got 0).
    End Content Validation: file:/Users/rchen/Desktop/bin.dat
        1 product validation(s) completed
Summary:
  1 error(s)
  0 warning(s)
  Product Validation Summary:
    0          product(s) passed
    1          product(s) failed
    0          product(s) skipped
  Referential Integrity Check Summary:
    0          check(s) passed
    0          check(s) failed
    0          check(s) skipped
  Message Types:
    1            error.table.records_mismatch
End of Report
Completed execution in 3992 ms

Where did 0 records come from? I found this when trying to validate something else, which I've also attached. x.xml, y.xml, and z.xml all point to x.dat, and they all have the same error.
[Archive.zip](https://github.com/NASA-PDS/validate/files/4604906/Archive.zip)

radiosci:
Validate 1.23.0 returns bogus error message about record count when checking binary tables.  In the example below, table 4 does have 9 records leaving 0 records not accounted for.  The error message is:
ERROR  [error.table.records_mismatch]   table 4: Number of records read is not equal to the defined number of records in the label (expected 9, got 0).
I can provide the PDS4 data file (23 MB) and label on request (unable to attach them here).

msbentley:
To add to this @jordanpadams and @hhlee445 I tested the 1.24.0 snapshot on a FITS file with a single ASCII table, and I get incorrect number of record warnings because the FITS has to have extra padding (to the end of the 2880 byte block) which validate sees as an extra record.

It's a warning, so not catastrophic, but is going to cause some users confusion!

jordanpadams:
@msbentley yeah i'm not sure what we should do here... on one hand we don't want to signal a false failure, but on the other hand, we want users to know when the record count doesn't match what we are reading out of the file. if you have some suggestions for a better way to handle this, we are open to trying something different

PR #261, opened 2020-10-30
add new pds-validate scripts

### LINKED ISSUE #259: rename validate CLI tool to pds-validate
Will duplicate scripts for backwards compatibility

jordanpadams:
starts towards PDS tools consistency to enable more usable access to components

resolves #259

### DIFF
--- a/src/main/assembly/tar-assembly.xml
+++ b/src/main/assembly/tar-assembly.xml
@@ -62,6 +62,9 @@ POSSIBILITY OF SUCH DAMAGE.
         <include>validate</include>
         <include>validate.bat</include>
         <include>validate-bundle</include>
+        <include>pds-validate</include>
+        <include>pds-validate.bat</include>
+        <include>pds-validate-bundle</include>
       </includes>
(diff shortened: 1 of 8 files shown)


jordanpadams:
@tloubrieu-jpl what do you think of this change? do you think it is worth it to do this for harvest and registry-mgr before we deploy? Or is it not worth it?

tloubrieu-jpl:
@jordanpadams we can do that, would that be ok regarding I&T since Richard already started to test the tools ?

jordanpadams:
@tloubrieu-jpl i guess that is my concern. either we do it now, or toss out the idea altogether (people will be annoyed if we pivot later). this is definitely a nice-to-have so maybe it isn't worth it?

tloubrieu-jpl:
@jordanpadams since it also requires to revise all the documentation I prefer that we give up that for this build. I would very much like to have that but maybe we can think of that consistent naming again when we have a uniform api client which would do everything we want. May be the client could include non API client features such as harvest or registry manager, that would be like a node package...

PR #267, opened 2020-11-16
Add logic to skip schematron validation on --skip-product-validation

### LINKED ISSUE #254: validate does not perform expediently when doing bundle-level validation against large bundles
**Describe the bug**
When invoking `validate` against a large bundle (i.e. 100k+ of files), and explicitly skipping product-validation, validate still takes hours to days to complete (estimated 2 days at a rate of 2 files per second being evaluated). 

According to @jordanpadams - `validate` should run in the time frame of "minutes" typically when running without product validation, though it hasn't been tested against bundles containing hundreds of thousands of products.

**To Reproduce**
1. Download the latest InSight bundle: https://pds-imaging.jpl.nasa.gov/data/nsyt/insight_cameras/
2. Run `validate` with skipping of product-level validation: 
```
validate -R pds4.bundle --skip-product-validation -t insight_cameras/ --report-file validate-tool-run-report-noprodcheck.txt
```

3. Notice `validate` will be scanning every XML file at a rate of about 2 per second. This means a bundle of 100k+ files will take days to complete, which is largely the same as when running WITH product-level validation.

**Expected behavior**
After step 2 above, expect `validate` to return within a few minutes at most. 

Possible solutions:
- Include an option to check only non-product XML files
- Include parallelization support to speed up XML file processing

** Version of Software Used**
`1.24.0`

**Desktop (please complete the following information):**
 - OS: RHEL 7

qchaupds:
resolves #254: validate does not perform expediently when doing bundle-level validation against large bundles

The reduction in time is due to not performing the time consuming process of validating each label file (associated with a data file) against the 9 schematrons if the --skip-product-validation is used.

Running with the --skip-product-validation should take about 20 minutes for the bundle uploaded by the user.

validate -R pds4.bundle -r /data/home/pds4/qchau/test_artifacts/large_bundle_with_skip_product_validation.json     --skip-product-validation -s json -t /data/home/pds4/insight_cameras/ > & /data/home/pds4/qchau/test_artifacts/log_with_skip_product_validation.txt

The test below without the --skip-product-validation flag can be ran as well but it may take a long time.   At the time as of this pull request, the test is still running:

validate -R pds4.bundle -r /data/home/pds4/qchau/test_artifacts/large_bundle_without_skip_product_validation.json                            -s json -t /data/home/pds4/insight_cameras/ > & /data/home/pds4/qchau/test_artifacts/log_without_skip_product_validation.txt

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/label/LabelValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/label/LabelValidator.java
@@ -33,6 +33,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -62,6 +63,8 @@ import javax.xml.xpath.XPathFactory;
 
 import net.sf.saxon.om.DocumentInfo;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.w3c.dom.Attr;
 import org.w3c.dom.CharacterData;
 import org.w3c.dom.Comment;
@@ -90,6 +93,7 @@ import org.xml.sax.helpers.AttributesImpl;
  *
  */
 public class LabelValidator {
+  private static final Logger LOG = LoggerFactory.getLogger(LabelValidator.class);
   private Map<String, Boolean> configurations = new HashMap<String, Boolean>();
   private List<URL> userSchemaFiles;
   private List<URL> userSchematronFiles;
@@ -118,6 +122,23 @@ public class LabelValidator {
   private Schema validatingSchema;
   private SchematronTransformer schematronTransformer;
   private XPathFactory xPathFactory;
+
+  private long filesProcessed = 0;
+  private double totalTimeElapsed = 0.0;
+
+  /**
+   * Returns the number of files processed by the validation function. 
+   */
+  public long getFilesProcessed() {
+      return(this.filesProcessed);
+  }
+
+  /**
+   * Returns the duration it took to run the validation function. 
+   */
+  public double getTotalTimeElapsed() {
+      return(this.totalTimeElapsed);
+  }
 
   /**
    * Default constructor.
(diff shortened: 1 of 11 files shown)


src/main/java/gov/nasa/pds/tools/label/LabelValidator.java
```diff
@@ -312,6 +333,26 @@ public synchronized void validate(ProblemHandler handler, URL url)
...
+      // Function return true if the validation against the schematron should be done or not.
+      // Note: schematron validation can be time consuming.  Only the bundle or collection should be validated against schematron.
+      // If the flag skipProductValidation is true, the labels belong to data files should not be done.
+      String nameOnly = Paths.get(url.getPath()).getFileName().toString();
+      Boolean validateAgainstSchematronFlag;
+      if (!this.skipProductValidation) {
+          validateAgainstSchematronFlag = true;
+      } else {
+          // If skip product validation is true, perform schematron validation bundle/collection labels only.
+          if (nameOnly.contains("bundle") || nameOnly.contains("collection")) {
```
jordanpadams: @qchaupds it is not required that a bundle or collection contain that information in the file name. if we want to check that, we need to actually read the XML and determine if it it's root is `Product_Bundle` or `Product_Collection`. depending upon the objects available in this class, you may be able to check what type of object this is? not sure what the easiest way to do this is.

src/main/java/gov/nasa/pds/tools/label/LocationValidator.java
```diff
@@ -226,8 +226,10 @@ public void validate(ValidateProblemHandler problemHandler, URL url) {
 			ruleContext.setCrawler(crawler);
 			ruleContext.setRule(rule);
 
+            LOG.info("validate:Submitting task {} to taskManager location {} rule {} ",task,location, rule.getCaption());
```
jordanpadams: @qchaupds can we make this a DEBUG message? we don't want to overload the INFO with even more information that is more for operational use.

src/main/java/gov/nasa/pds/tools/label/LocationValidator.java
```diff
@@ -226,8 +226,10 @@ public void validate(ValidateProblemHandler problemHandler, URL url) {
...
             LOG.debug("validate:Submitting task to taskManager location {} rule {} ",location, rule.getCaption());
             taskManager.submit(task);
+            LOG.info("validate:Returning from task {} to taskManager location {} rule {} ",task,location, rule.getCaption());
```
jordanpadams: @qchaupds can we also make this a DEBUG message?

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/CollectionReferentialIntegrityRule.java
```diff
@@ -56,8 +59,12 @@
   
   private static final String LOGICAL_IDENTIFIER =
       "//*[starts-with(name(),'Identification_Area')]/logical_identifier";
+
+  private static final String VERSION_ID =
```
jordanpadams: @qchaupds this looks to be unused in this class

jordanpadams (changes requested):
looks good! a few minor updates recommended

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/CollectionReferentialIntegrityRule.java
qchaupds: The unused variable has been removed for next commit.

src/main/java/gov/nasa/pds/tools/label/LocationValidator.java
qchaupds: The info logging has been removed for next commit.

src/main/java/gov/nasa/pds/tools/label/LocationValidator.java
qchaupds: The info logging has been removed for next commit.

src/main/java/gov/nasa/pds/tools/label/LabelValidator.java
qchaupds: Working on the fix to inspect the file itself to see what type it is.  Will be ready for next commit.

jordanpadams (approved):

PR #277, opened 2020-12-15
Throw an error if the ASCII String containing the '+' symbol and ASCI…

### LINKED ISSUE #11: Update allowable field_format values per Standards Reference definition regarding [+|-] characters
Initial issue identified:

> Line 105 of the attached .xml has
>   <field_format>%+8s</field_format>
> even though data_type is ASCII_STRING, and page 29 of the Standards says, "In PDS4 labels, the "+" is forbidden for string fields."
> 
> OK, I apologize that this is so ticky-tack, but the real data file had -, and the data was not left-justified, and validate flagged that to my surprise. So this is your penalty for catching something so minute in the first place.

Conversation between @mcayanan and @rjoyner

> I definitely agree that the constraints listed below should be validated.  Just not sure SCH is the most efficient ?  In the past we have discussed that Validate validate stuff that couldn’t be (easily) validate in SCH ?
> Let’s discuss…
>  
> [+|-]
> Either a "+" or "-", but never both. The "-" may be used for string fields, to indicate that the string is (or should be) left-justified in the field. This is actually the preferred way to present most string values in character tables, so the field_format value for fields with a data type of ASCII_String will nearly always begin with a "-". Similarly for any of the date/time type fields. In PDS4 labels, the "-" prefix is forbidden for all numeric fields (integers, floating point numbers, and numbers using scientific notation).
> The "+" may be used with numeric fields to indicate that an explicit sign is included in the field for input, and should be displayed on output. In PDS4 labels, the "+" is forbidden for string fields.
>  
> (1)    SCH to ensure -- Either a "+" or "-", but never both.
>  
> SCH to ensure the constraints would be enforced in accordance with the  <data_type>. 
>  
> (2)    So, I think all of the following <data_type>  values would NOT allow “+“:
>  
> ASCII_Short_String_Collapsed
> ASCII_Short_String_Preserved
> ASCII_String
> ASCII_Text_Collapsed
> ASCII_Text_Preserved
>  
>  
> (3)    All of the following <data_type> values would NOT allow “-“:
>  
> ASCII_NonNegative_Integer
> ASCII_Numeric_Base16
> ASCII_Numeric_Base2
> ASCII_Numeric_Base8
> ASCII_Real
> ASCII_Integer
>  
> ('ComplexLSB16', 'ComplexLSB8', 'ComplexMSB16', 'ComplexMSB8', 'IEEE754LSBDouble', 'IEEE754LSBSingle', 'IEEE754MSBDouble', 'IEEE754MSBSingle', 'SignedBitString', 'SignedByte', 'SignedLSB2', 'SignedLSB4', 'SignedLSB8', 'SignedMSB2', 'SignedMSB4', 'SignedMSB8', 'UnsignedBitString', 'UnsignedByte', 'UnsignedLSB2', 'UnsignedLSB4', 'UnsignedLSB8', 'UnsignedMSB2', 'UnsignedMSB4', 'UnsignedMSB8')">
>  
> The (1) & (2) are fairly simple SCH to write.  The (3) not so much…

qchaupds:
Throw error when field format contains '+' for ASCII String and '-' for ASCII Number fields
 
Closes NASA-PDS/validate#11
 
The issue is validate did not flag a bad ASCII_String format [%+8s].  The reference manual states that ASCII_String and a few other related should be left justified.
With the new code, it does now flag it.
The code also flag any ASCII number format if they contain the minus symbol as in [%-2d]

Tested in DEV:

validate -R pds4.label  -r report_github11_bad_1.json -s json -t src/test/resources/github11/test_data/science_index_bad_1.xml

validate -R pds4.label  -r report_github11_bad_2.json -s json -t src/test/resources/github11/test_data/science_index_bad_2.xml

validate -R pds4.label --skip-context-validation -r report_github11_good.json -s json -t src/test/resources/github11/test_data/science_index_good.xml

The first 2 tests should fail with 3 errors each: 2 for context and 1 for bad format.
The last test with the --skip-context-validation should validate successfully of the format.

If the last test was ran without the --skip-context-validation, there will be two errors for context.   It was left in since it may be confusing to have a failed test
for a valid input.


src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -84,9 +94,40 @@
...
+      ASCII_NUMBER_TYPE_LIST.add("ASCII_Numeric_Base16".toUpperCase());
+      ASCII_NUMBER_TYPE_LIST.add("ASCII_Numeric_Base2".toUpperCase());
+      ASCII_NUMBER_TYPE_LIST.add("ASCII_Numeric_Base8".toUpperCase());
+      ASCII_NUMBER_TYPE_LIST.add("ASCII_Real".toUpperCase());
+      ASCII_NUMBER_TYPE_LIST.add("ASCII_Integer".toUpperCase());
+  }
+
+  private static String PRODUCT_OBSERVATIONAL = "Product_Observational";
+  private static String TABLE_CHARACTER = "Product_Observational/File_Area_Observational/Table_Character";
+  private static String RECORD_CHARACTER = "Product_Observational/File_Area_Observational/Table_Character/Record_Character";
```
jordanpadams: @qchaupds could we use a shorter absolute XPath here instead of including this nesting? that way we do not have to update the code whenever a new product is introduced that uses it. e.g.

```
//Table_Character
//Table_Character/Record_Character
```

thoughts?

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -133,11 +174,136 @@ public boolean isApplicable(String location) {
...
+
+  private void validateAsciiStringFieldsFormat(List<String> fieldFormatList, List<String> fieldTypeList, List<String> fieldNumberList) {
+    // ASCII String related fields should not have '+' in the format.
+
+    int indexToList = 0;
+    // Loop through all elements fieldFormatList.  If field is in ASCII_STRING_TYPE_LIST, check if prohibited '+' is in the format.
+    for (String oneFieldFormat : fieldFormatList) {
+        //LOG.debug("validateAsciiStringFieldsFormat:indexToList,oneFieldFormat,fieldTypeList.get(indexToList)  {},{},{}",indexToList,oneFieldFormat,fieldTypeList.get(indexToList));
+        // Check for prohibited plus symbol '+' in format field : %+8s
+        if (this.ASCII_STRING_TYPE_LIST.contains(fieldTypeList.get(indexToList).toUpperCase())) {
```
jordanpadams: `ASCII_STRING_TYPE_LIST` is a static variable, so we shouldn't be using `this.` here.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -133,11 +174,136 @@ public boolean isApplicable(String location) {
...
+
+  private void validateAsciiNumberFieldsFormat(List<String> fieldFormatList, List<String> fieldTypeList, List<String> fieldNumberList) {
+    // ASCII Number related fields should not have '-' in the format.
+
+    int indexToList = 0;
+    // Loop through all elements fieldFormatList.  If field is in ASCII_NUMBER_TYPE_LIST, check if prohibited '-' is in the format.
+    for (String oneFieldFormat : fieldFormatList) {
+        //LOG.debug("validateAsciiNumberFieldsFormat:indexToList,oneFieldFormat,fieldTypeList.get(indexToList)  {},{},{}",indexToList,oneFieldFormat,fieldTypeList.get(indexToList));
+        // Check for prohibited minus symbol '-' in format field : %-2d
+        if (this.ASCII_NUMBER_TYPE_LIST.contains(fieldTypeList.get(indexToList).toUpperCase())) {
```
jordanpadams: `ASCII_NUMBER_TYPE_LIST` is a static variable, so we shouldn't be using `this.` here.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -133,11 +174,136 @@ public boolean isApplicable(String location) {
...
+     // Example of an invalid field_format for ASCII_String (right justified):
+     //          <data_type>ASCII_String</data_type>
+     //          <field_length unit="byte">8</field_length>
+     //          <field_format>%+8s</field_format>
+     //          <description>Number of the asteroid.</description>
+     int numFields = 0;
+
+     try {
+        XMLExtractor extractor = new XMLExtractor(getTarget());
+        TinyNodeImpl productObservationalNode = extractor.getNodeFromDoc(this.PRODUCT_OBSERVATIONAL);
```
jordanpadams: looks like static variables are being accessed with `this` throughout. can we just do a quick scrub of this new code to make sure we are accessing this properly.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -133,11 +174,136 @@ public boolean isApplicable(String location) {
...
+    int indexToList = 0;
+    // Loop through all elements fieldFormatList.  If field is in ASCII_NUMBER_TYPE_LIST, check if prohibited '-' is in the format.
+    for (String oneFieldFormat : fieldFormatList) {
+        //LOG.debug("validateAsciiNumberFieldsFormat:indexToList,oneFieldFormat,fieldTypeList.get(indexToList)  {},{},{}",indexToList,oneFieldFormat,fieldTypeList.get(indexToList));
+        // Check for prohibited minus symbol '-' in format field : %-2d
+        if (this.ASCII_NUMBER_TYPE_LIST.contains(fieldTypeList.get(indexToList).toUpperCase())) {
+            if (oneFieldFormat.contains("%-")) {
+                LOG.error("ASCII Number related fields should not contain '-' symbol in field_format [" + oneFieldFormat + "] in field_number " + fieldNumberList.get(indexToList));
+                addTableProblem(ExceptionType.ERROR, 
+                    ProblemType.TABLE_FILE_READ_ERROR,
```
jordanpadams: I think we should probably use the `ProblemType.INVALID_LABEL` since this is really a bad label specification, not the data itself. does that seem reasonable?

jordanpadams (changes requested):
@qchaupds functionality looks great and works as expected. commented with a few things that need to updated.

also, I think this code improvement should go in the `TableFieldDefinitionRule` class since what we are really catching here is errors in the definition of table fields, not the content of the data itself. I could be convinced otherwise, but just a thought.

thanks for the great work!

jordanpadams (approved):
@qchaupds looks great!

note: i updated the test data to update/remove those invalid context references so we no longer have those erroneous other errors to worry about.

PR #284, opened 2021-01-14
Improve logic when checking registered product version id against the…

### LINKED ISSUE #278: Registered context products file does not retain older versions of context products
The file **registered_context_products.json** that comes with Validate does not list all versions of context products. It lists only the latest versions. Therefore products that refer to the older versions are now invalid, according to Validate, which reports the error "error.label.context_ref_not_found". 

**To Reproduce**
1. Download a document from the InSight RISE document collection: https://pds-geosciences.wustl.edu/insight/urn-nasa-pds-insight_documents/document_rise/trk-2-34-revn-l5_tnf.xml 
2. Run Validate version 1.24.0 on this product, e.g. "validate trk-2-34-revn-l5_tnf.xml -R pds4.label -v1"
3. Observe resulting error message "ERROR  [error.label.context_ref_not_found]   line 34: 'Context product not found: urn:nasa:pds:context:investigation:mission.insight::1.0"
4. Observe that **urn:nasa:pds:context:investigation:mission.insight::1.0** is in fact a valid context product listed at https://pds.nasa.gov/data/pds4/context-pds4/investigation/mission.insight_1.0.xml. There is now a version 2.0 of this context product, but version 1.0 is still registered. 
5. Observe that **registered_context_products.json** lists only version 2.0 of this context product.

**Expected behavior**
I expected **registered_context_products.json** to include all versions of context products, so that archives that had previously passed validation would continue to pass.

** Version of Software Used**
Validate 1.24.0

**Desktop (please complete the following information):**
 - OS: Windows 10.0.19042 
 - Browser: Chrome
 - Version: 87.0.4280.88

qchaupds:
Improve logic when checking registered product version id against the provided lidvid in label.
 
Closes NASA-PDS/validate#278
 
The issue is the registered product only keep the latest version.
When validate run, it does an exact look up, which make previous version invalid.
The code fix was to make sure the LID matches and the provided version id is less or equal to 
registered product.

Tested in DEV:

% validate -R pds4.label -r report_github278_label_invalid.json -s json -t src/test/resources/github278/invalid/trk-2-34-revn-l5_tnf_invalid.xml

% validate -R pds4.label -r report_github278_label_valid.json   -s json -t src/test/resources/github278/valid/trk-2-34-revn-l5_tnf.xml

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/ContextProductReferenceValidationRule.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/ContextProductReferenceValidationRule.java
@@ -44,6 +44,9 @@ import org.w3c.dom.Document;
 import org.w3c.dom.NodeList;
 import org.w3c.dom.Node;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import gov.nasa.pds.tools.label.ExceptionType;
 import gov.nasa.pds.tools.label.SourceLocation;
 import gov.nasa.pds.tools.util.ContextProductReference;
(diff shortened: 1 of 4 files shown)


jordanpadams (approved):
looks great @qchaupds !

Issue #290, opened 2021-01-27
Migrate subset of existing regression tests to cucumber behavioral testing


jordanpadams:
Let's migrate tests from this build to the new framework started in #282. here are a list of issues we can use for reference, but we do not need to create any new tests. Let's just migrate those that already have regression tests.

* #281 
* #278 
* #6 
* #256 
* #257 
* #246 
* #273 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

**Applicable requirements**
:unicorn: #1
:unicorn: #2

tloubrieu-jpl:
@qchaupds  has issue with integration of cucumber test in maven.

Issue #240, opened 2020-06-19
Unexpected error for data collection in a sub-directory


jordanpadams:
**Describe the bug**
Validate throws an error for `unallowed directory name` even though it is the base data collection.

**To Reproduce**
1. download bundle from `pds-gamma:/data/local/starbase/data/pds4/test-bundles/pds4-compil-comet-v1.0/`
2. run validate with the pds4.bundle rule enabled

**Expected behavior**
This should validate successfully because it is a data collection. This is a misinterpretation of the PDS4 Standard Section 2B.2.2 and 6C.2.3, which calls this out as an example, not a requirement. We should just throw a WARNING message here:

```
WARNING  [warn.sub_directory.unallowed_name]   Directory name is reserved and not in the recommended location in the bundle root directory.
```

Applicable requirements:
* 🦄 #239

qchaupds:
As Jordan had pointed out, currently if an unexpected sub directory is found outside of the root directory, it thinks it is an error:
and report UNALLOWED_BUNDLE_SUBDIRECTORY_NAME in src/main/java/gov/nasa/pds/tools/validate/rule/pds4/PDS4Problems.java

`public static final ProblemDefinition UNALLOWED_BUNDLE_SUBDIRECTORY_NAME = new ProblemDefinition(
            ExceptionType.ERROR,
            ProblemType.UNALLOWED_BUNDLE_SUBDIR_NAME,
            "Directory name is not allowed except in the bundle root directory",
            Standard.PDS4_STANDARDS_REFERENCE,
            "6C.2.2"
    );`

So a preferred ExceptionType is a WARNING with a different message:

`public static final ProblemDefinition UNALLOWED_BUNDLE_SUBDIRECTORY_NAME = new ProblemDefinition(
            ExceptionType.WARNING,
            ProblemType.UNALLOWED_BUNDLE_SUBDIR_NAME,
            "Directory name is reserved and not in the recommended location in the bundle root directory",
            Standard.PDS4_STANDARDS_REFERENCE,
            "6C.2.2"
    );`

PR #293, opened 2021-01-28
Implement initial framework to be able to start developing tests and …

### LINKED ISSUE #282: Implement initial framework to be able to start developing tests
This ticket is intended to develop the initial framework for tests and how to create a new one. Each test should contain at least 1 success-oriented test case and at least 1 expected failure test case.

qchaupds:
Implement initial framework to be able to start developing tests and improve error handling for data collection in a sub-directory

Closes NASA-PDS/validate#282 NASA-PDS/validate#283 NASA-PDS/validate#240 and NASA-PDS/validate#290

There is an existing integration test that gets run automatically.
This new capability allows tester/developer to add new tests for validation with the new cucumber framework
without having to change Java code.

Tested in DEV:

To test NASA-PDS/validate#282:

The test code for cucumber behavior testing will need to be compiled with:

% mvn package -DskipTests

The -DskipTests is to skip the (normal) existing integration tests.

A file will be created target/validate-1.25.0-SNAPSHOT-bin.zip by the command above and need to be unzipped with:

% cd target
% unzip validate-1.25.0-SNAPSHOT-bin.zip

A directory of .jar files will be created in validate-1.25.0-SNAPSHOT/lib

To run the cucumber testing assuming the version value in pom.xml for artifactId 'validate' is '1.25.0-SNAPSHOT'

% cd ../
% java -cp "target/test-classes:target/validate-1.25.0-SNAPSHOT/lib/*" io.cucumber.core.cli.Main target/test-classes/features

12 Scenarios (12 passed)
48 Steps (48 passed)
0m23.997s

To test NASA-PDS/validate#240:

% validate -R pds4.bundle  -r report_github240_bundle_valid.json   -s json -t src/test/resources/github240/valid/bundle_kaguya_derived.xml

% validate -R pds4.bundle  -r report_github240_bundle_invalid.json -s json -t src/test/resources/github240/invalid/bundle_kaguya_derived.xml


jordanpadams:
@tloubrieu-jpl could you take a look at this? we may want to discuss if this is how we want to implement the cucumber testing or if we want to integrate in maven build?

jordanpadams (changes requested):
@qchaupds I think I would prefer if we ran these cucumber tests within a JUnit Runner class so they actually execute during the maven test phase. if we have any tests that are duplicated within the `ValidationIntegrationTests` let's just remove them.

@tloubrieu-jpl does that seem reasonable to you?

jordanpadams (changes requested):
@qchaupds thanks for tryin to dig up a solution here. but we really needed this to work with `mvn test`. I was able to track down in the JUnit subsection under https://cucumber.io/docs/cucumber/api/?sbsearch=junit#running-cucumber, that Cucumber is based off of JUnit 4 and we are using JUnit 5. Because of this, we needed to add `junit-vintage-engine`. I then added a `CucumberTest` JUnit runner class. And it now appears to run nominally with `mvn test`. However, a few tests are failing. can you please take a look?

jordanpadams (changes requested):
@qchaupds so after poking around at this a bit, i can't figure out why the caching problem is still happening.

that being said, i don't think we should spend any more time debugging it. i've disabled ValidationIntegrationTests for now, and let's just migrate all of them to cucumber. once we fix the caching issue in validate next build, we can try to re-enable the tests to verify the fix.

thoughts?

PR #295, opened 2021-02-11
Make ROOT element check namespace-aware

sxntxn:
This patch makes the SCHEMATRON assert:
```
"The ROOT element must be one of the allowed types."
```
namespace-aware.

E.g., the root-element declaration:
```
<pds:Product_Observational xmlns:pds="http://pds.nasa.gov/pds4/pds/v1" >
```
is semantically equivalent to:
```
<Product_Observational xmlns="http://pds.nasa.gov/pds4/pds/v1" >
```
... so it should also be accepted as correct.

Please let me know what you think.


jordanpadams:
@sxntxn thanks so much for the PR! unfortunately, we cannot update the schematron to do this check because that is part of the [PDS4 Information Model](https://github.com/NASA-PDS/pds4-information-model), and that is managed separate from this repo. We actually should probably remove those from validate because we don't even use them anymore in the code :-) .

feel free to propose a change in that other repo and we can get the fix included. additionally, if a more immediate fix is required, we may be able to update the code where the schematron is read in by the Java code and see if that helps.

does that make sense? can you send over an example label you would like to check so we can include that in the ticket?

sxntxn:
Thanks!

I've filed a bug report with the other repository:
https://github.com/NASA-PDS/pds4-information-model/issues/315

Issue #294, opened 2021-02-09
Content validation incorrectly reports error for floating-point values out of specified min/max range


sslavney:
**Describe the bug**
With content validation turned on, Validate incorrectly issues the error message "error.table.field_value_out_of_min_max_range" for some floating-point fields in a CSV table, where Field_Statistics/maximum and Field_Statistics/minimum are defined. 

For example, given a field with this definition:

                <Field_Delimited>
                    <name>sample_time_increment</name>
                    <field_number>126</field_number>
                    <data_type>ASCII_Real</data_type>
                    <description>Time increment covered by each sample in this RDR-processed sounding</description>
                    <Field_Statistics>
                        <local_identifier>sample_time_increment_stats</local_identifier>
                        <maximum>0.12345</maximum>
                        <minimum>0.12345</minimum>
                    </Field_Statistics>
                </Field_Delimited>

Validate gives this error message:
      ERROR  [error.table.field_value_out_of_min_max_range]   table 1, record 357, field 126: Field has a value '0.12345' that is greater than the defined maximum value '0.12345'. 

There are two things going on here. 

First, apparently the internal representation of either the field maximum or the actual value has too much precision, e.g. 0.1234999999999 or something like that, which makes the comparison meaningless.

Second, Field_Statistics/minimum and Field_Statistics/maximum merely report the statistics for the value found in that data product, if I am reading the definitions correctly. They don't mean that values outside that range are invalid. So the message should be a warning, not an error. There's nothing wrong with the data; it's the label that needs to be fixed.

**To Reproduce**
Using the attached data file and label minmax-error.*, run Validate as follows:
validate minmax-error.xml -R pds4.label -v2 -r validate-minmax-error.txt --skip-context-validation
Notice the many error messages like the one in the example, apparently all for floating-point values (although I didn't do an exhaustive search).

**Expected behavior**
In cases like the example, where the actual value (0.12345) is within the expected min-max range, no error nor warning message should be issued.
In cases where the actual value is above Field_Statistics/maximum or below Field_Statistics/minimum, a warning message should be reported. It could be something like "Field has a value "nnn" that is greater than the expected maximum 'mmm'.
 
** Version of Software Used**
1.24.0

**Test Data / Additional context**
See attached label, data file, and validation report.

**Desktop (please complete the following information):**
 - OS: Windows 10

 - Browser: Chrome
 - Version: irrelevant
[...]

Issue #292, opened 2021-01-28
CCB-264: Make the Line Feed (LF) character an allowed record delimiter


jordanpadams:
**Link to SCR:** https://pds-jira.jpl.nasa.gov/browse/CCB-264

**SCR Details:** [CCB-264.pdf](https://github.com/NASA-PDS/validate/files/5884205/CCB-264.pdf)

See nasa-pds/pds4-information-model#299 for what has been updated in the information model.

Test data: 
[minimal_test_product_line_endings.zip](https://github.com/NASA-PDS/validate/files/5884209/minimal_test_product_line_endings.zip)

jordanpadams:
📆 february status: delayed due to issues with new testing framework. on schedule for B11.1.

jordanpadams:
resolved per #296

PR #296, opened 2021-02-16
Updates per #291 #292 #294

### LINKED ISSUE #291: When validating a product with a bad schematron definition, bundle validation also fails indicating the associated product does not exist
**Describe the bug**
When I run validate against a bundle with a product that has invalid schematron definitions, it looks like validate is include that product in the bundle validation for referential integrity.

**To Reproduce**
1. when i run validate like this:

```
bin/validate -R pds4.bundle -t $TEST_DATA_PATH/test-bundles/pds4-compil-comet-v1.0/bundle.xml
```

2. one of the warning messages you get for product validation is:

```
  PASS: file:/data/local/starbase/data/pds4/test-bundles/pds4-compil-comet-v1.0/phys_char/document/pccref.xml
      WARNING  [warning.label.bad_schematypens]   Could not find expected pattern [schematypens=] in label file:/data/local/starbase/data/pds4/test-bundles/pds4-compil-comet-v1.0/phys_char/document/pccref.xml
      WARNING  [warning.label.missing_schematron_spec]   No schematrons specified in the label
        53 product validation(s) completed
```

3. And then later on during the bundle validation portion you get this erroneous warning:

```
  PASS: file:/data/local/starbase/data/pds4/test-bundles/pds4-compil-comet-v1.0/phys_char/document/pccref.txt
      WARNING  [warning.file.not_referenced_in_label]   File is not referenced by any label
        46 integrity check(s) completed
```

the label does in fact refer to this product:
```
            <Document_File>
                <file_name>pccref.txt</file_name>
                <document_standard_id>7-Bit ASCII Text</document_standard_id>
            </Document_File>
```

**Expected behavior**
if I fix the typo in the schematron

**Version of Software Used**
```
gov.nasa.pds:validate
Version 1.25.0-SNAPSHOT
Release Date: 2021-01-27 20:19:12

All rights reserved.
```
**Test Data / Additional context**
Can be found on pds-gamma under `$TEST_DATA_PATH/test-bundles/pds4-compil-comet-v1.0/`

qchaupds:
Allows Line Feed (LF) character to terminate a record in table format file and resolves associated file even if schematron is not available in a label.

Closes NASA-PDS/validate#291, NASA-PDS/validate#292 and NASA-PDS/validate#294

This pull request resolves three issues:

1.  If the schematron is not available, the label parsing and validation still take place,
2. The line feed character is now allowed to end a record,
3. The comparisons of floating points is corrected when the min and max values are the same as the record content.

Tested in DEV:

To test NASA-PDS/validate#291

% validate -R pds4.bundle  -r report_github291_bundle_valid.json   -s json -t src/test/resources/github291/valid/bundle_kaguya_derived.xml
% validate -R pds4.bundle  -r report_github291_bundle_invalid.json -s json -t src/test/resources/github291/invalid/bundle_kaguya_derived.xml

To test NASA-PDS/validate#292:

Test a valid label, one with Line Feed (LF) and one with Carriage Return/Line Feed (CR/LF) record terminator.

% validate --skip-context-validation -R pds4.label  -r report_github292_label_valid_lf.json     -s json -t src/test/resources/github292/valid/minimal_test_product_lf.xml
% validate --skip-context-validation -R pds4.label  -r report_github292_label_valid_crlf.json   -s json -t src/test/resources/github292/valid/minimal_test_product_crlf.xml

Test a invalid label, one with LF and one with CR/LF record terminator.  The error will be because the line length specified (999) is too long (60 actual length).

% validate --skip-context-validation -R pds4.label  -r report_github292_label_invalid_lf.json   -s json -t src/test/resources/github292/invalid/minimal_test_product_lf.xml
% validate --skip-context-validation -R pds4.label  -r report_github292_label_invalid_crlf.json -s json -t src/test/resources/github292/invalid/minimal_test_product_crlf.xml

To test NASA-PDS/validate#294:

% validate --skip-context-validation -R pds4.label -r report_github294_bundle_invalid.json -s json -t src/test/resources/github294/invalid/minmax-error.xml
% validate --skip-context-validation -R pds4.label -r report_github294_bundle_valid.json   -s json -t src/test/resources/github294/valid/minmax-error.xml


src/main/java/gov/nasa/pds/tools/validate/rule/pds4/FileReferenceValidationRule.java
```diff
@@ -37,6 +37,7 @@
 
 import gov.nasa.pds.tools.label.ExceptionType;
 import gov.nasa.pds.tools.util.FileSizesUtil;
+import gov.nasa.pds.tools.util.FileReferencedMapList;
```
jordanpadams, line 40: @qchaupds looks like you are missing this class in your commit?

jordanpadams (changes requested):

jordanpadams (changes requested):
@qchaupds I thought I sent this in our email discussions, but can't track it down.

#292 requires we validate against an unreleased version of the IM (1.16.0.0 aka 1G00). i've added those to the repo.

additionally, validate needs to validate according to the line delimiter specified in `//record_delimiter`. for example, for a `Table_Character`:
```
    <Table_Character>
      <name>Test Instrument Table Data</name>
      <offset unit="byte">0</offset>
      <records>4</records>
      <description>This table is a Character table</description>
      <record_delimiter>Carriage-Return Line-Feed</record_delimiter>
```
vs.
```
    <Table_Character>
      <name>Test Instrument Table Data</name>
      <offset unit="byte">0</offset>
      <records>4</records>
      <description>This table is a Character table</description>
      <record_delimiter>Line-Feed</record_delimiter>
```
This will need to apply to all tables.

i've updated the test data to reflect the new IM version, and the Line-Feed for those test products. however, I think the software will need to be updated to key off of that `//record_delimiter` attribute value.

here is how you can run the software to specify schema/schematrons via command-line:

```
validate-1.25.0-SNAPSHOT/bin/validate -t src/test/resources/github292/valid/minimal_test_product_crlf.xml \
    --schema  src/test/resources/github292/PDS4_PDS_1G00.xsd \
    --schematron src/test/resources/github292/PDS4_PDS_1G00.sch \
    --skip-context-validation
```

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -426,14 +447,23 @@ public void validateTableDataContents() throws MalformedURLException,
           } else {  // We have either a character or delimited table
             boolean manuallyParseRecord = false;
             String line = reader.readNextLine();
+            int lineNumber = 0;
             while (line != null) {
               progressCounter();
+              lineNumber += 1;
               
               if (!line.endsWith("\r\n")) {
```
jordanpadams: @qchaupds per #292, I think the logic can be much simpler here. most importantly, per this ticket, the check for line ending must be keyed off of the value in the `record_delimiter` attribute. here is some pseudocode for a simple check:
```
// Very similar to the original logic in this method, 
// except we first read the value of the <record_delimiter> attribute
if record_delimiter.value() == "Carriage-Return Line-Feed" AND !line.endsWith("\r\n"):
    throw ERROR
// Next, we need to do an else if because CRLF ends in a new line so we need to do some additional checks
else if record_delimiter.value() == "Line-Feed":
    if (line.endsWith("\r\n") OR !line.endsWith("\n"):
        throw an ERROR
```

Per the sample data, here is an example  of where you can find `record_delimiter`:
```
  <File_Area_Observational>
    <File>
      <file_name>minimal_test_product_crlf.tab</file_name>
      <file_size unit="byte">240</file_size>
      <md5_checksum>5bc560bb3554b8d705ec36e3905726f5</md5_checksum>
    </File>

    <Table_Character>
      <name>Test Instrument Table Data</name>
      <offset unit="byte">0</offset>
      <records>4</records>
      <description>This table is a Character table</description>
      <record_delimiter>Carriage-Return Line-Feed</record_delimiter>
```

jordanpadams (changes requested):

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
jordanpadams: additionally, I don't think we care about the ASCII value. you should just be able to just check if the string ends with `\r\n` or just `\n`. Let me know if that makes sense? or we can setup some time tomorrow to discuss

src/main/java/gov/nasa/pds/tools/util/TabulatedUtil.java
```diff
@@ -0,0 +1,63 @@
...
+      // Get the value of 'record_delimiter' in a label.  The returned value can be null.
+      // <Table_Character>
+      //     <record_delimiter>Carriage-Return Line-Feed</record_delimiter>
+
+      String recordDelimiter = null;
+
+      List<TinyNodeImpl> xmlModels = new ArrayList<TinyNodeImpl>();
+      try {
+          SAXSource saxSource = new SAXSource(Utility.openConnection(getTarget()));
+          DocumentInfo docInfo = LabelParser.parse(saxSource); // Parses a label.
```
jordanpadams, line 47: @qchaupds apologies for poking at this more but I've come across a few more issues

---

- [x] **issue 1**:

we have to read in the label to DOM for other aspects of content validation as well. for instance:

1. here: https://github.com/NASA-PDS/validate/blob/master/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java#L125
2. and here: https://github.com/NASA-PDS/validate/blob/master/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java#L125 https://github.com/NASA-PDS/validate/blob/master/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java#L167

I'm not sure if the read on L125 can be set as a global or not, but if not, at minimum we should refactor the code with L167 so we read in the XML object and can use it for your check + the remainder of the code.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -451,28 +436,51 @@ public void validateTableDataContents() throws MalformedURLException,
...
+                        reader.getCurrentRow());
+                  }
               } else {
-                addTableProblem(ExceptionType.DEBUG,
-                    ProblemType.CRLF_DETECTED,
-                    "Record ends in carriage-return line feed.",
-                    dataFile,
-                    tableIndex,
-                    reader.getCurrentRow());              
+                  // If cannot find a record delimiter, check for the default carriage return line feed.
```
jordanpadams, line 475: @qchaupds if record_delimiter is null or not in ("Carriage-Return Line-Feed", "Line-Feed") this is an invalid label. schematron validation will throw an error, but we can just add another error here so we fail gracefully on content validation as well.

src/main/java/gov/nasa/pds/tools/util/TabulatedUtil.java
```diff
@@ -0,0 +1,63 @@
...
+          XMLExtractor extractor = new XMLExtractor(docInfo);
+          xmlModels = extractor.getNodesFromDoc("Product_Observational/File_Area_Observational/Table_Character");
```
jordanpadams: - [x] 1. if we could use the `Xpaths` class to define these XPaths as variables that would great. see https://github.com/NASA-PDS/validate/blob/master/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java#L128 for an example of usage. hard coding in these classes can be really hard to track down in the future.

- [x] 2. This does not just apply to `Table_Character`, this applies to all places where `record_delimiter` exists in a product. if possible, I would just generalize this to `//record_delimiter`. if that is not possible, then we will need to find all the objects this applies to in the [PDS4 IM Spec](https://pds.nasa.gov/datastandards/documents/im/v1/index_1G00.html)

jordanpadams (changes requested):

jordanpadams (changes requested):
## Review
We are getting closer. The test you have look good. I just updated the data a bit for the invalid tests to be more CRLF v LF-centric to make sure they are testing against one another.
 
I then added a whole bundle other test data for other types of tables that are included with this update to the IM. More specifically:
•	Table_Delimited
•	Stream_Text.
•	Inventory
 
For Table_Delimited, this should be similar implementation to Table_Character, you will just need to handle that use case. I have hoping that since Inventory extends Table_Delimited, once  you make the update to work for the Table_Delimited test cases, it will work with Inventory. If that is the case, this will make our lives a lot easier for all the other possible extensions of these tables. If that is not the case and we should discuss our options.
 
For Stream_Text, I wouldn’t worry too much about this right now. We do not test the content of these products right now, and that is not the end of the world. We have another ticket in the backlog we will get to to address this.

Additionally, the DDWG just noted that if a file specifies a LF in the label, but it ends in a CRLF, we should catch this and throw an WARNING since it is unlikely this was intended. A new test label, `src/test/resources/github292/table_character/invalid/minimal_test_product_lf_2.xml` was created to test this. Right now, validate says this is valid. However, even though the records are 60 characters long (valid) and end in a `\n` (valid), technically there is a CRLF at the end of the line, so we should catch this and throw a WARNING.

## Work Items
Here is a checklist of items that need to be done:

- [x] update tests to use new directory structure

See checklist tagged in the code for additional items.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -426,23 +431,55 @@ public void validateTableDataContents() throws MalformedURLException,
...
+              if (recordDelimiter != null) {
+                  // Check for how the line ends keying off what was provided in the label.
+                  // If the delimiter is "Carriage-Return Line-Feed" then the line should end with a carriage return and a line feed.
+                  if (recordDelimiter.equalsIgnoreCase("Carriage-Return Line-Feed") && !line.endsWith("\r\n")) {
+                    addTableProblem(ExceptionType.ERROR,
+                        ProblemType.MISSING_CRLF,
+                        "Record does not end in carriage-return line feed.",
+                        dataFile, tableIndex, reader.getCurrentRow());
+                    manuallyParseRecord = true;
+                  } else if (recordDelimiter.equalsIgnoreCase("Line-Feed") && !line.endsWith("\n")) {
```
jordanpadams: @qchaupds per my review comments, we need to improve this logic.

- [x] update logic for handling CRLF line endings when LF is specified in the label

```
if recordDelimiter == Line-Feed:
   if !line.endsWith("\n"): THROW ERROR   // this is your current error thrown
   if line.endsWith("\r\n"): THROW ERROR   // this is a new error. label says LF, but line ends in CRLF
```

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
```diff
@@ -340,6 +344,7 @@ public void validateTableDataContents() throws MalformedURLException,
         } else {
           LOG.debug("table instanceof TableCharacter: else");
           TableCharacter tc = (TableCharacter) table;
```
jordanpadams, line 347: @qchaupds per comments in review, we need to add additional logic in the code about for handling record delimiters for `Table_Delimited`:

- [x] update code and add tests for table_delimited

right now we only handle this for `Table_Character` per your updates below.

jordanpadams (changes requested):

src/main/java/gov/nasa/pds/tools/util/TabulatedUtil.java
qchaupds: As it turns out, another class contains a function to fetch the delimiter already so there's no need to re-parse the label file.

jordanpadams:
@qchaupds excellent work! thanks so much for dealing with a moving target here.

PR #305, opened 2021-03-05
Error handle table fields with double quotes within beginning and end…

### LINKED ISSUE #298: validate misses double quotes within a delimited table
Line 5 of the attached .csv has
  1998_1228,1998_1228-0,"The data sets consist of (1) three small ""photometric spot"" sub-framed ...

That is illegal based on the PDS4 Standards, section 4C.1, bullet 6:
  Fields may optionally be bracketed by a pair of bounding double quotes... If bracketing double quotes are used, there may be no double quotes within the field itself.

validate passes the file with no errors. PDS4 viewer correctly chokes on it.
[Archive.zip](https://github.com/NASA-PDS/validate/files/5998735/Archive.zip)

qchaupds:
Handle double quotes in delimited table

Closes NASA-PDS/validate#298

This pull request resolves issue if a double quote occurs inside a field that starts with a double quote in a tabulated file. 
Now the code reports the error in the report and flag it as an ERROR.

Tested in DEV:

% validate -R pds4.label --skip-context-validation -r report_github298_label_invalid.json   -s json -t src/test/resources/github298/invalid/sentences.xml

% validate -R pds4.label --skip-context-validation -r report_github298_label_valid.json   -s json -t src/test/resources/github298/valid/sentences.xml


jordanpadams:
@qchaupds it looks like there are still a lot of code changed in this PR and I'm not entirely sure what all of it does or why we have all of it? would you mind take a quick scrub through here and clean this up a bit and remove unused code / classes if they exist.

also, it still looks like we are using this "smartSplit" method, but I'm not sure exactly what the complication is here?  it seems like the logic should be pretty straightforward for this check now that PDS4-JParser is updated.

```
read field
if field has "bracketing surrounding quotes":
    raise error if there are any quotes used inside
end
remove "bracketing surround quotes"  // so we can proceed with the rest of validation like before
```

is there some other complex logic we need here that I am missing?

Issue #327, opened 2021-04-13
validate fails to process large data file


msbentley:
## 🐛 Describe the bug
When trying to validate a large (4.8 GB) data file, validate does not crash, but appears to show no output - after 15 minutes no "dots" or other output are displayed, but CPU usage is still high.

## 📜 To Reproduce
1. Try to validate (with content validation) a large file

## 🕵️ Expected behavior
Some indication of progress is given (this is especially important for large files, and the chunk size must be appropriate to the expected duration of the validation)

## 📚 Version of Software Used
validate 2.0.2

## 🩺 Test Data / Additional context
Test data is proprietary and too large to upload here, but can be provided offline if required.

## 🖥 System Info
 - OS: linux, ubuntu, 20.04

---

Most likely related to https://github.com/NASA-PDS/pds4-jparser/pull/31/files

qchaupds:
@msbentley Mark, can you tell us the file extension of the file in question so we can target our development?  The issue was fixed for some products but not all back in November of 2020.
The one we are currently creating a test file for is .csv and it will take some time to create as there isn't one that large in our test artifacts.  The one created yesterday is not a true .csv file as only the first 5 records had a carriage/line feed, the rest were random data and to perform a validation for CSV, we need to have all records comply.

msbentley:
Hi @qchaupds - this is actually a rather complicated file, which we are trying to discourage, but at the moment this is what we have - the label describes 26322 (!) `Table_Character` tables each containing 120 records, each with a group field repetition (160 times) of a single field. The rationale was to try to encode a cube in ASCII, I believe. It is only a secondary data format, since the primary data are in an array, but the team wanted an ASCII alternative as well (and since arrays can only use binary types...). So we have 26k frames from a 120x160 detector, encoded as ASCII_Integer.

I could hazard a guess that it is the number of `Table_Character` instances that is throwing validate, rather than the sheer size of the file...?

qchaupds:
@msbentley Hi Mark, can you tell use the file extension you are using?  Is it ".csv" or ".tab"?  Validate uses the file extension to use the correct validator.  The one we are trying out now is a .csv file (4GB) and is having good success so far.

qchaupds:
Also @msbentley, inorder for us to test the file you have, we may need a copy of it on our system.

We can try to test with files we create but won't know until we try on the actual file it is failing for you.

qchaupds:
[...]

Issue #326, opened 2021-04-12
File-size check fails for large data files


msbentley:
## 🐛 Describe the bug
Trying to validate a large product, but skipping data validation, the new file size check (label vs actual file size) apparantly fails when the data file is too large, with the error:

```
ERROR  [error.label.filesize_mismatch]   Generated filesize '-1' does not match supplied filesize '5060141280' in the product label for 'file:/home/mbentley/Desktop/dictionary_test/mertis/mer_raw_sc_tis_20201214_4_table.dat'
```
As you can see the file in question is quite large (4.8GB).

## 📜 To Reproduce
1. Try to validate a large data file with validate 2.0.2

## 🕵️ Expected behavior
File-size checks do not fail for large files ;-)

## 📚 Version of Software Used
Version 2.0.2

## 🖥 System Info
 - OS: linux, ubuntu, 20.04

---

** :unicorn: Applicable requirements**

qchaupds:
@msbentley Hi Mark, with some small changes in validate module, we were able to validate a very large file (with the --skip-content-validation flag)

{pds-dev3.jpl.nasa.gov}/data/home/pds4/qchau/test_artifacts/github326 132 % ls -lctr
total 4194320
-rw-r--r-- 1 qchau pds        405 Apr 21 09:17 spectra_data_collection_inventory.csv
-rw-r--r-- 1 qchau pds 4294967296 Apr 21 09:18 gentoo_root.csv
-rw-r--r-- 1 qchau pds 4294967701 Apr 21 09:20 very_large_file.csv
-rw-r--r-- 1 qchau pds       4490 Apr 21 09:22 spectra_data_collection_inventory.xml

The file very_large_file.csv was created by combining the spectra_data_collection_inventory.csv with 5 records and the large file gentoo_root.csv.   We can test the file_size validation but not the content since validate would fail with such a large file.

{pds-dev3.jpl.nasa.gov}/data/home/pds4/qchau/test_artifacts/github326 136 % egrep "file_name|file_size" spectra_data_collection_inventory.xml
      <file_name>very_large_file.csv</file_name>
      <file_size unit="byte">4294967701</file_size>


{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 131 % validate -R pds4.label --skip-content-validation --skip-context-validation -r report_github326_label_valid.json -s json -t /data/home/pds4/qchau/test_artifacts/github326/spectra_data_collection_inventory.xml

[...]

PR #331, opened 2021-04-27
Raise a warning for non-compliant PDF files and fix file_size validat…

### LINKED ISSUE #164: As a user, I want to validate PDF files are PDF/A
Per section 8A.2.2 of the PDS4 SR, PDF files must be PDF/A-1a or PDF/A-1b.

> Each document must be saved in one or more PDS-compliant formats using the Product_Document class. PDS-compliant formats for documents are “flat UTF-8” text, PDF/A1a (which is preferred), or PDF/A-1b. 

https://pdfbox.apache.org/1.8/cookbook/pdfavalidation.html

---
UPDATE
After some digging and tracking down some resources from https://pdfa.org, it looks like veraPDF may be a better solution and it actually wraps PDFBox for us. It also appears to have active development ongoing, which is great as PDF continues to evolve:
 
https://verapdf.org/software/
https://docs.verapdf.org/develop/
https://github.com/veraPDF/veraPDF-library

qchaupds:
…ion for very large files

Please merge pds4-jparser first (from the other pull request) as validate depends on pds4-jparser.

Closes NASA-PDS/validate#164, NASA-PDS/validate#326, and NASA-PDS/validate#327

This pull request resolves three issues: Raise a warning if a PDF file is not PDF/A compliant and fixes file_size validation for very large files. 

Tested in DEV for #164:

% validate -R pds4.label --skip-context-validation -r report_github164_label_pdfa_invalid.json -s json  -t src/test/resources/github164/invalid/document_test_1_pdf.xml
% validate -R pds4.label --skip-context-validation -r report_github164_label_pdfa_valid.json -s json  -t src/test/resources/github164/valid/document_pdfa_valid.xml

Tested in DEV for #326 and #327 (Note that no regression was added for this issue since the file is too big to be added to the git repository):
The file is over 3GB in size and contains 38 million records.

% validate -R pds4.label --skip-content-validation --skip-context-validation  -r report_github326_label_valid_big_file_check_only.json -s json -t /data/home/pds4/qchau/test_artifacts/github326/spectra_data_collection_inventory.xml

% validate -R pds4.label --skip-context-validation -r report_github326_label_valid_big.json -s json -t /data/home/pds4/qchau/test_artifacts/github326/spectra_data_collection_inventory.xml

% validate -R pds4.label --skip-context-validation -r report_github326_label_valid_small.json -s json -t /data/home/pds4/qchau/test_artifacts/github326/spectra_data_collection_inventory_small.xml

% validate -R pds4.label --skip-context-validation -r report_github326_label_invalid.json -s json -t /data/home/pds4/qchau/test_artifacts/github326/spectra_data_collection_inventory_invalid.xml


src/main/java/gov/nasa/pds/tools/util/MD5Checksum.java
```diff
@@ -54,23 +62,49 @@ public static String getMD5Checksum(URL url) throws Exception {
...
+
+        if (bufferSize > inChannel.size()) {
+            bufferSize = (int) inChannel.size();
         }
-      } while (bytesRead != -1);
-      return md5.digest();
-    } finally {
-      input.close();
-    }
+        ByteBuffer byteBuffer = ByteBuffer.allocate(bufferSize);
```
jordanpadams, line 79: @qchaupds similar to my comment in pds4-jparser, do we think it may make sense to group a lot of these data content checks together into one "walk" through the bits of the file? this seems like a 3rd time we are now stepping through the product to check something, which is a 3x performance degradation for very large files. it may not be possible, but worth a discussion.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/FileReferenceValidationRule.java
```diff
@@ -220,16 +231,45 @@ private boolean validate(DocumentInfo xml) {
...
+                            ExceptionType.WARNING, ProblemType.NON_PDFA_FILE,
+                            urlRef.toString() + " is not valid PDF/A file or does not exist");
+                        problems.add(new ValidationProblem(def, target,
+                            fileObject.getLineNumber(), -1));
+                        passFlag = false;
+                    }
+                }
               } else if ("md5_checksum".equals(child.getLocalPart())) {
                 checksum = child.getStringValue();
+LOG.debug("FileReferenceValidationRule:validate:checksum {}",checksum);
```
jordanpadams: @qchaupds very nitpicky here, but can we fix this whitespace if we want to keep these log messages in this method?

jordanpadams (changes requested):

src/main/java/gov/nasa/pds/tools/util/MD5Checksum.java
jordanpadams: from @qchaupds The MD5Checksum.java is a standalone utility to calculate the checksum.  It has no knowledge about the file type or record size. It should also be noted that if the `--skip-content-validation` is used, the checksum and file_size do get validated.   The checksum check is apart from the content validation.

jordanpadams: will table this re-design for a later refactor

jordanpadams (approved):

PR #347, opened 2021-05-18
Check that all internal references are valid references to other prod…

### LINKED ISSUE #308: As a user, I want to check that all Internal References are valid references to other PDS4 products within the current validating bundle
1. For more information on how to populate this new feature request, see the PDS Wiki on User Story Development: https://github.com/NASA-PDS/nasa-pds.github.io/wiki/Issue-Tracking#user-story-development

2. Do the best you can with template. If it is too difficult to create a "story" just jot down as much info as you can.

## Motivation
...so that I can ensure referential integrity of references within the bundle to other products within the same parent bundle

## Additional Details
Need to confirm that every LID or LIDVID referenced in an Internal_Reference class exists.

Concatentate every LID/LIDVID in Identification Area from every label in a bundle. Then for each xml label, verify that each LID/LIDVID not in the Identification_Area is included in the concatenated list. If not giving a warning that there is a missing product. Ideally also check against all registered LIDs & LIDVIDs.

Check with Richard Chen, he has a python script that does this checking.

## Acceptance Criteria
**Given** a product that contains one or more `Internal_Reference`s to product LID/LIDVIDs within the same parent bundle
**When I perform** validation of the bundle
**Then I expect** to validate that all LIDs/LIDVIDs to products within the bundle are valid references

qchaupds:
…ucts within the current bundle

Should check that all internal references are valid references to other products within the current bundle.
If the logical_identifier of theproduct containing the said reference does NOT belong to the bundle, the WARNING about
non-existent identifier is not raised but other WARNINGs may be raised, perhaps not being a member of any collection.

Closes NASA-PDS/validate#308

Tested in DEV for #308

**There should be no WARNING regarding a reference pointing to a logical identifier.**
% validate -R pds4.bundle  -r report_github308_bundle_valid.json   -s json -t src/test/resources/github308/valid/bundle_kaguya_derived.xml

**There should be 2 warnings for referencing a logical identifier for a product not found in this bundle for label bundle_kaguya_derived.xml**
**There should be 1 warning for referencing a logical identifier for a product not found in this bundle for label kgrs_calibrated_spectra_per1.xml**

% validate -R pds4.bundle  -r report_github308_bundle_invalid.json   -s json -t src/test/resources/github308/invalid/bundle_kaguya_derived.xml


src/main/java/gov/nasa/pds/tools/util/LabelUtil.java
```diff
@@ -214,6 +223,84 @@ public static String getIMVersion(DOMSource source, URL context) {
...
+     return(lidOrLidVidReferences);
+  }
+
+  /**
+   * Get the local identifiers the label (as a DOMSource)
+   * @param source The content of context as a DOMSource.
+   # @param context The location of the label being parsed from.
+   * @return logicalIdentifiers A list of logical identifiers in this label.
+   */
+  public static ArrayList<String> getLogicalIdentifiers(DOMSource source, URL context) {
```
jordanpadams, line 301: @qchaupds seems like this method is almost identical to the `getLidVidReferences` method above. can we refactor this logic?

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/BundleReferentialIntegrityRule.java
```diff
@@ -88,12 +99,263 @@ public boolean isApplicable(String location) {
...
+                }
+            }
+            indexToFilenames += 1;
+        } // end for loop
+      } catch (Exception e) {
+          e.printStackTrace();
+      }
+  }
+
+  private boolean hasReferenceIDAndFilenameComboAdded(String singleLidorLidVidReference, URL filename) {
```
jordanpadams, line 216: @qchaupds can we use some other data structure here instead of combining this info with a `slash`? this seems like we are trying to force this into a string, but you never know if we will need to utilize these data structures for something else in the future.

additionally, is there any way we can use the `lidvid` of the product in order to do these mappings instead of the filename? the filename will most likely work, but, the LIDVID is really the identifier for the product and avoids any possible weird collisions.

jordanpadams (changes requested):
@qchaupds looks good. a couple changes requested.

additionally, for future issues, we need to think about our test cases a bit. i appreciate the diligence in providing a valid/invalid test case for each bug fix, but our test data is getting huge (~62M). we can start with whatever data we need to perform our testing, but if the test data already exists, or another test already accomplishes one of our test cases, we should just call that out in the PR. if it isn't obvious that another test case meets our needs, we can totally just add more data. not a huge deal. just something to keep in mind.

for instance, we already have this kaguya data elsewhere in the test suite, and we definitely already have a valid bundle run somewhere in here. we can just re-use that test data, or even those test cases, in the pull request.

src/main/java/gov/nasa/pds/tools/util/LabelUtil.java
qchaupds: The 2 functions have now been refactored into a 3rd common function.
Will be reflected when code is checked in.

Noted on your comment about adding new valid test when an existing one could work already.  The only disadvantage is if a particular test resource is removed in the future, check should be made to see if other tests in the feature file is not re-using it.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/BundleReferentialIntegrityRule.java
qchaupds: The slash has been removed to avoid confusion.  We are trying to compare a combination of two strings so a simple concatenation was sufficient.

The mapping is now from a lid_reference or lidvid_refence to a logical_identifier (instead of from a filename).

jordanpadams (approved):
looks good. thanks @qchaupds !

Issue #257, opened 2020-10-21
Product with incorrect table binary definition pass validation


fran-vallejo:
**Describe the bug**
A product with an invalid table definition which was correctly detected in validate 1.22.2, pass as valid with the current version in master (1.25.0-SNAPSHOT).

**To Reproduce**
Validate attached test product with validate 1.22.2 and current master version (1.25.0-SNAPSHOT).

**Expected behavior**
validate should report the test product attached as invalid as the table field definition is not valid. 

** Version of Software Used**
validate 1.22.2 and repository head of master (1.25.0-SNAPSHOT).

Note that the product defines a Table_Binary, but as field_length is mandatory for each field, I'm not sure if this is the expected behaviour or a regression. If behaviour is expected, please disregard this ticket. 

**Test Data / Additional context**
[testProduct.zip](https://github.com/NASA-PDS/validate/files/5415297/testProduct.zip)

Execution with validate 1.22.2:
```
BepiVirtualBox:~/Software$ validate-1.22.2/bin/validate -t /home/fvallejo/git/pds4-packager-clean/pds4-validate/src/test/resources/pds4/mer_raw_sc_tir_20181113_01.xml -add-context-products /home/fvallejo/git/pds4-packager-clean/pds4-validate/src/test/resources/pds4-validate-data/local_context_products.json 

PDS Validate Tool Report

Configuration:
   Version                       1.22.2
   Date                          2020-10-21T09:26:43Z

Parameters:
   Targets                       [file:/home/fvallejo/git/pds4-packager-clean/pds4-validate/src/test/resources/pds4/mer_raw_sc_tir_20181113_01.xml]
   Severity Level                WARNING
   Recurse Directories           true
   File Filters Used             [*.xml, *.XML]
   Data Content Validation       on
   Product Level Validation      on
   Max Errors                    100000
   Registered Contexts File      /home/fvallejo/Software/validate-1.22.2/resources/registered_context_products.json
   Non Registered Contexts File  /home/fvallejo/git/pds4-packager-clean/pds4-validate/src/test/resources/pds4-validate-data/local_context_products.json



Product Level Validation Results

  PASS: gov.nasa.pds.validate.ValidateLauncher
      WARNING  [warning.product_not_registered]   Non-registered context products should only be used during archive development. All context products must be registered for a valid, released archive bundle. 
        1 product validation(s) completed

[...]

PR #393, opened 2021-08-26
Check Header of File_Area_Ancillary and provide meaningful error mess…

### LINKED ISSUE #361: validate does not check Header of a File_Area_Ancillary nor does not provide a meaningful error message for an incorrect Table_Character offset
## 🐛 Describe the bug
When fine tuning the generation of the label for a table file, I first struggled to debug my label because the error message I was obtaining was not pointing me towards the error. If the Table_Character offset attribute is incorrect validate interprets that there is only one record of length 1 (see attached validate_offset.log file).

Ultimately the error was that the offset was 1 byte off. 

While debugging this error, I played around with the length of the Header attribute of the Table_Character and I found out that the object_length of the header is not checked and if an incorrect value (in my case it should correspond to the Table_Character offset minus one byte) is provided, the label validates.

## 📜 To Reproduce
The original tab file is attached (maven_orb_rec_210101_210401_v1.orb) along with the correct label (maven_orb_rec_210101_210401_v1.xml), a label with an incorrect header length that validates (maven_orb_rec_210101_210401_v1.xml.header) and a label with the incorrect offset that fails to validate (maven_orb_rec_210101_210401_v1.xml.offset).

The command used is:
$validate --local -t maven_orb_rec_210101_210401_v1.xml

## 🕵️ Expected behavior
I would expect an error message if the header length is incorrect.
I would expect a more meaningful error message if the offset is incorrect.

## 📚 Version of Software Used
Version 2.0.3

## 🩺 Test Data / Additional context
Test data is attached. Note that .txt has been added to the non .log extensions.

[maven_orb_rec_210101_210401_v1.orb.txt](https://github.com/NASA-PDS/validate/files/6659613/maven_orb_rec_210101_210401_v1.orb.txt)
[maven_orb_rec_210101_210401_v1.xml.header.txt](https://github.com/NASA-PDS/validate/files/6659614/maven_orb_rec_210101_210401_v1.xml.header.txt)
[maven_orb_rec_210101_210401_v1.xml.offset.txt](https://github.com/NASA-PDS/validate/files/6659615/maven_orb_rec_210101_210401_v1.xml.offset.txt)
[maven_orb_rec_210101_210401_v1.xml.txt](https://github.com/NASA-PDS/validate/files/6659616/maven_orb_rec_210101_210401_v1.xml.txt)
[validate_offset.log](https://github.com/NASA-PDS/validate/files/6659617/validate_offset.log)

## 🖥 System Info
 - OS: macOS Catalina 10.15.7

---

qchaupds:
…ages when the offset overlap between Header and table data or table fields overlap between each others.

1. Add test resources :src/test/resources/github361
2. Add meaningful error message when fields overlap :src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
3. Add logic to check for overlap of header total length with table offset, and add meaningful error message when offset is incorrect :src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
4. Modify logic to read Table_Character tags from all possible main tags not just Product_Observational :src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableFieldDefinitionRule.java

Refs:

https://github.com/nasa-pds/validate/issues/361 validate does not check Header of a File_Area_Ancillary nor does not provide a meaningful error message for an incorrect Table_Character offset

Closes https://github.com/nasa-pds/validate/issues/361 validate does not check Header of a File_Area_Ancillary nor does not provide a meaningful error message for an incorrect Table_Character offset

The error messages weren't as meaningful to the user so extra verbiage were added so the user can make the correction in the label.
Also, previously, the code was not reading the Table_Character tag but it does now.
The reason was the Table_Character can also be File_Area_Ancillary product.

Since issue #361 is a follow on issue #257, a new test set was added to allow the tester to validate the meaningful-ness of the error messages.
No new regression tests were added.

Tested in DEV:

Case 1:  The offsets and headers are good.

```
% validate -R pds4.label --skip-context-reference-check -r report_github361_label_valid_mini_good_offset.json  -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.good.xml
```
Case 2: The table offset is too early before the header total length ends.
```
% validate -R pds4.label --skip-context-reference-check -r report_github361_label_invalid_mini_offset_too_early.json  -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.too.early.xml >& t11
```
Case 3: The table offset is past the header total length (which is OK) but throws off table validation since the fields are all wrong because
the first field is starting in the middle of the 1st record and not the first byte.
```
% validate -R pds4.label --skip-context-reference-check -r report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation.json  -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.xml
```
Case 4 :  The start position of field 2 is too early and overlaps with field 1
```
% validate -R pds4.label --skip-context-reference-check -r report_github361_label_invalid_mini_header.json  -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.invalid.minimal.header.xml
```

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
@@ -251,10 +251,12 @@ public class FieldValueValidator {
             // if (((i+1)<fields.length) && (fields[i].getOffset()+fields[i].getLength() > fields[i+1].getOffset()))
             // The next line is hard to read.  Perhaps the parenthesis should surround the 2nd > comparison.
         	if (((i+1)<fields.length) && (fields[i].getOffset()+fields[i].getLength()) > fields[i+1].getOffset()) {
+                int currentFieldEndsAt = fields[i].getOffset()+fields[i].getLength();
+                int nextOffsetShouldBe = fields[i].getOffset()+fields[i].getLength() + 1;
         		String message = "This field overlaps the next field. Current field ends at " 
-        				+ (fields[i].getOffset()+fields[i].getLength()) 
-        				+ ". Next field starts at " + fields[i+1].getOffset();
-                LOG.error(message);
+        				+ currentFieldEndsAt  
+        				+ ". Next field starts at " + fields[i+1].getOffset() + " but should be at least at " + nextOffsetShouldBe;
+                LOG.error("{}","MESSAGE_1:" + message);
         		addTableProblem(ExceptionType.ERROR,
         				ProblemType.FIELD_VALUE_OVERLAP,
         				message,
@@ -295,10 +297,13 @@ public class FieldValueValidator {
                 // issue_257: Product with incorrect table binary definition pass validation
                 // Corrected logic: using the OR logic || and put parenthesis surround the 2nd check for readability.
         		if ((fields[i].getOffset()>fields[i+1].getOffset()) || (fields[i].getOffset()+fields[i].getLength()) > fields[i+1].getOffset()) {       
+                    int currentFieldEndsAt = fields[i].getOffset()+fields[i].getLength();
+                    int nextOffsetShouldBe = fields[i].getOffset()+fields[i].getLength() + 1;
         			String message = "This field overlaps the next field. Current field ends at " 
-        					+ (fields[i].getOffset()+fields[i].getLength()+1) 
-        					+ ". Next field starts at " + (fields[i+1].getOffset()+1);
-                    LOG.error("{}",message);
+        					//+ (fields[i].getOffset()+fields[i].getLength()+1) 
+        					+ currentFieldEndsAt
+        					+ ". Next field starts at " + (fields[i+1].getOffset()+1) + " but should be at least at " + nextOffsetShouldBe;
+                    LOG.error("{}","MESSAGE_2:" + message);
         			addTableProblem(ExceptionType.ERROR,
         					ProblemType.FIELD_VALUE_OVERLAP,
         					message,
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableDataContentValidationRule.java
@@ -585,6 +585,48 @@ public class TableDataContentValidationRule extends AbstractValidationRule {
       return(tableIsLineOrientedFlag);
   }
 
+  private void performOffsetCheck(long readerOffset, int headerOffset, int headerSize, URL dataFile, int tableIndex, String tableName) {
+          // Add new checks for the header and the table data area to see if they overlap.
+          //
+          // Case 1:
+          // This case is OK because the table starts after the header ends.
+          // Validate only know that the table offset starts after the header ends.
+          // It cannot know if the offset will NOT cause any problem with the parsing of each record.
+          // That is the responsibility of the user to make any correction if it turns out that
+          // many fields are invalid due to a "bad" offset.
+          //
+          //     [this is the header]
+          //                          [this is the table]
+          //
+          // Case 2:
+          // This is NOT OK because the table starts before the header ends.
+          // The header and the table overlaps.
+          //
+          //     [this is the header]
+          //           [this is the table]
+          //
+          // reader.getOffset()        = from beginning of file to table data area
+          // headerOffset + headerSize = total header length
+
+          if (readerOffset < (headerOffset + headerSize)) {
+              // Case 2:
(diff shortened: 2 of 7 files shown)


jordanpadams (changes requested):
@qchaupds a couple related issues here:

- [ ] 1. offset does not appear to be read or used in content validation

when I change the offset of your product that fails to 500, the product validates successfully. however, that should put the reader somewhere in the middle of the data file, which should throw an error. it seems like validate may just be using the header offset + object_length to start reading, when it should simply always be using the offset for where to start in the file.

- [ ] 2. the header / offset overlap check must apply to all Table_*
when I change your test product to use Table_Binary instead of Table_Character, it does not properly throw the error message you are referring to.

qchaupds:
@jordanpadams Issue 1: I am not able to see the same issue you see.  Can you be more specific on the label you are using?

Added a new label: src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.500.xml
where the offset is changed from 300 to 500 and the error can be observed (because it threw everything off):

This test fail because the offset is in the middle of the table.

```
% validate -R pds4.label --skip-context-reference-check -r report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation_500.json  -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.500.xml >& t12

{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 110 % egrep "status|label" report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation_500.json
    "ruleType": "pds4.label",
      "status": "FAIL",
      "label": "file:/home/qchau/sandbox/validate/src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.500.xml",
```

This also fail because it is in the middle of the table (with the offset of 300)

```
{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 112 % validate -R pds4.label --skip-context-reference-check -r report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation.json  -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.xml >& t12

{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 113 %  egrep "status|label" report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation.json
    "ruleType": "pds4.label",
      "status": "FAIL",
      "label": "file:/home/qchau/sandbox/validate/src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.xml",
{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 114 % 
```

Can you clarify if the offset is late (beyond the header) is an error because it is beyond the header (You had mentioned that it was OK in the past because the user wanted to skip more after the header) or because the parsing is off because the fields don't align anymore with their expected formats?  Currently, the code is depending on the parsing to fail to report the error.

qchaupds:
@jordanpadams My bad.  I just realized a file is missing in github.  Stay tuned.  Don't know what happened.

qchaupds:
@jordanpadams Please grab latest commit 7ebff46 and have another review.

The following shows validate for the label with "Table_Binary":
```
validate -R pds4.label --skip-context-reference-check -r report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation_table_binary.json -s json -t src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.table.binary.xml > & t12

{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 110 % egrep "status|label"  report_github361_label_invalid_mini_offset_late_is_ok_but_throws_of_table_validation_table_binary.json            
    "ruleType": "pds4.label",
      "status": "FAIL",
      "label": "file:/home/qchau/sandbox/validate/src/test/resources/github361/maven_orb_rec_210101_210401_v1.minimal.offset.late.is.ok.but.throws.off.table.validation.table.binary.xml",
```
Both issues should be resolved now.

jordanpadams (approved):
@qchaupds looks good! very weird when I set the offset in `maven_orb_rec_210101_210401_v1.minimal.offset.too.early.xml` to be basically 268-270, it still validates successfully. but i guess that is just a coincidence that those 3 bytes don't throw off the validation of the rows/fields.

PR #398, opened 2021-09-01
Fix bug when checking for offset for fields that are not of BIT type

### LINKED ISSUE #392: Validate throws incorrect overlap error when first Field_Bit has length 1
## 📜 To Reproduce
2. Type 'validate test1.xml'
3. See error

## 🖥 System Info
 - MacBook Pro 2   OS 11.1
 - Safari Version 14.0.2 (16610.3.7.1.9)
[field_bit_error.tar.gz](https://github.com/NASA-PDS/validate/files/7048542/field_bit_error.tar.gz)

---

qchaupds:
1. Add test resource : src/test/resources/github392
2. Add logic to check for data_type of field before making the offset checks : src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
3. Add 3 new tests : src/test/resources/features/validate.feature

Refs:

https://github.com/nasa-pds/validate/issues/392 Validate throws incorrect overlap error when first Field_Bit has length 1

Closes https://github.com/nasa-pds/validate/issues/392 Validate throws incorrect overlap error when first Field_Bit has length 1

The data_type UnsignedBitString requires some refinement in checking for offset of each field.  Validate must know
what the data_type is for a particular field before making the check.  The check in question is modified to only performing the normal
offset between one field to another if the data_type is NOT UnsignedBitString.

Tested in DEV:

Case 1:
Running validate against a valid label with with data_type UnsignedBitString.  No errors expected.

```
% validate -R pds4.label --skip-context-validation -r report_github392_label_valid.json -s json -t src/test/resources/github392/test1_valid.xml
```

Case 2: 
Running validate against a label with Table_Binary with data_type UnsignedBitString with two bit fields overlap.

```
% validate -R pds4.label --skip-context-validation -r report_github392_label_invalid_1.json -s json -t src/test/resources/github392/test1_invalid.xml
```

totalErrors = 1
   error.table.field_value_overlap 1
              "message": "The bit field overlaps the next field. Current stop_bit_location: 7. Next start_bit_location: 6"

Case 3: 
Running validate against a label with Table_Binary with data_type SignedMSB4 with two errors:

 1. Field 1 and field 2 overlaps
 2. The offset of 16 greater than the size of the record which is 7 bytes.

```
validate -R pds4.label --skip-context-validation -r report_github392_label_invalid_2.json -s json -t src/test/resources/github392/INVALID_odf07155_msgr_11.xml
```

totalErrors = 2
   error.table.field_value_overlap 1
             "message": "This field overlaps the next field. Current field ends at 4. Next field starts at 3 but should be at least at 5"

   error.table.bad_field_read 1
              "message": "Error while getting field value: Field offset '16' is greater than the number of bytes in the record '7'"


jordanpadams (approved):
@qchaupds looks good. I was thinking this may be something we add to pds4-jparser, but then looking at this, this is strictly validating the label contents, not the content of the file itself. thanks!

src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
```diff
@@ -170,11 +170,20 @@ public void validate(TableRecord record, FieldDescription[] fields, boolean chec
...
+    // if fieldIsBitStringFlag is false before checking for offset.
+    // The reason is the bit field can have the same offset as the next field because the fields are in bits.
+
     int actualFieldNumber = 1;
     for (int i = 0; i < fields.length; i++) {
+      boolean fieldIsBitStringFlag = false; // Flag to store whether field is a UNSIGNEDBITSTRING or not.
       String value = "dummy_value";   // Set to a dummy value to allow inspection when the value changed to a legitimate value.
       //LOG.info("validate:i,fields.length {},{}",i,fields.length);
       try {
+        if (fields[i].getType().toString().toUpperCase().contains("BIT")) {
```
jordanpadams, line 184: @qchaupds actually, sorry! could we use these enums within [pds4-jparser FieldType](https://github.com/NASA-PDS/pds4-jparser/blob/main/src/main/java/gov/nasa/pds/label/object/FieldType.java) class instead? 

```
	/** An unsigned bit string in a packed field. */
	UNSIGNEDBITSTRING("UnsignedBitString", new BitFieldAdapter(false), true),
```

radiosci:
If I understand the solution, Validate will no longer check for field overlap when any field has data type UnsignedBitString.  I don't think that's very responsive, and it will lead to undetected errors in Packed Data Fields.  I'd actually prefer to use the current version of Validate and know that it is buggy when the first field has length 1 bit.  
What happens if the field has data type SignedBitString?  Obviously the field in those cases will always have length 2 bits or more; but does this mean that SignedBitFields are checked while UnsignedBitFields are not?

qchaupds:
@radiosci Actually, validate will still check for field overlaps for Packed Data Fields.

Here's an example of a bad label where field 2 starts at 6 which overlaps field 1 which starts at 1 and ends at 7.

```
                               <field_number>1</field_number>
                               <field_location unit="byte">1</field_location>
                               <field_length unit="byte">1</field_length>
                                           <field_number>1</field_number>
                                           <start_bit_location>1</start_bit_location>
                                           <stop_bit_location>5</stop_bit_location>
                                           <field_number>2</field_number>
                                           <start_bit_location>6</start_bit_location>
                                           <stop_bit_location>8</stop_bit_location>
                               <field_number>2</field_number>
                               <field_location unit="byte">2</field_location>
                                     <field_length unit="byte">1</field_length>
                                                 **<field_number>1</field_number>
                                                 <start_bit_location>1</start_bit_location>
                                                 <stop_bit_location>7</stop_bit_location>
                                                 <field_number>2</field_number>
                                                 <start_bit_location>6</start_bit_location>
                                                 <stop_bit_location>8</stop_bit_location>**
                               <field_number>3</field_number>
                               <field_location unit="byte">3</field_location>
                                     <field_length unit="byte">1</field_length>
                                                 <field_number>1</field_number>
                                                 <start_bit_location>1</start_bit_location>
                                                 <stop_bit_location>1</stop_bit_location>
                                                 <field_number>2</field_number>
                                                 <start_bit_location>2</start_bit_location>
                                                 <stop_bit_location>8</stop_bit_location>
                                     <field_number>4</field_number>
                                     <field_location unit="byte">4</field_location>
                                     <field_length unit="byte">1</field_length>
                                                 <field_number>1</field_number>
                                                 <start_bit_location>1</start_bit_location>
                                                 <stop_bit_location>2</stop_bit_location>
                                                 <field_number>2</field_number>
                                                 <start_bit_location>3</start_bit_location>
                                                 <stop_bit_location>8</stop_bit_location>

```

Running validate 
```
{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 115 %validate -R pds4.label --skip-context-validation -r report_github392_label_invalid_1.json -s json -t src/test/resources/github392/test1_invalid.xml
```

will result in the following error:

```
          "dataFile": "file:/home/qchau/sandbox/validate/src/test/resources/github392/test1.dat",
          "messages": [
            {
              "severity": "ERROR",
              "type": "error.table.field_value_overlap",
              "table": 1,
              "record": 1,
              "field": 2,
              "message": "The bit field overlaps the next field. Current stop_bit_location: 7. Next start_bit_location: 6"
            }

```

PR #399, opened 2021-09-09
Use enum type defined in FieldType instead of hard-coded string

### LINKED ISSUE #392: Validate throws incorrect overlap error when first Field_Bit has length 1

qchaupds:
1. Use enum type defined in FieldType instead of hard-coded string : src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java

Refs:

https://github.com/nasa-pds/validate/issues/392 Validate throws incorrect overlap error when first Field_Bit has length 1

Closes https://github.com/nasa-pds/validate/issues/392 Validate throws incorrect overlap error when first Field_Bit has length 1

This pull request is a follow up on the previous pull request for the same issue.  The only change was to use an enum type instead of a hard-code string.  

All regression tests passed.

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
@@ -177,11 +177,12 @@ public class FieldValueValidator {
 
     int actualFieldNumber = 1;
     for (int i = 0; i < fields.length; i++) {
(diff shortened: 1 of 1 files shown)


jordanpadams (approved):

jordanpadams:
🎉

PR #418, opened 2021-10-15
Update builds to use pinned sandbox Roundup action

MJJoyce:
Point both of our builds at the updated sandbox Roundup action for
initial un-sandboxed tests.

Note, there's ongoing discussion about using `validate` as our Java test guinea pig. Hold off on merging until that is resolved!


jordanpadams:
@MJJoyce added the WIP label to this PR so it isn't picked up by our Slack notifications for stale PRs. once you think this is ready to merge, you can just remove that label and it will start nagging us.

MJJoyce:
:+1: Thanks @jordanpadams

PR #483, opened 2022-03-08
products are now distributed to "deinterlaced" partitions rather than block partitions

alexdunncs:
@jordanpadams per earlier discussion

## 🗒️ Summary
This minimizes the chance of workload disparity between partitions, for example where one partition contains only browses and another contains only large images, which reduces efficiency.

Testing on a Mars2020 MEDA bundle resulted in runtime reduction from 7min to 5min. 

N.B. this changes the suffix for per-partition files from a two-char alpha id to an integer id

## ⚙️ Test Data and/or Report
No tests exist for `validate-bundle`, that I can see

## ♻️ Related Issues
n/a


jordanpadams (approved):

jordanpadams:
yay community PRs 🎉 🎉 🎉

PR #525, opened 2022-06-27
Add getters/setters for context configs

### LINKED ISSUE #241: As a developer, I want an API method enable specifying of non-registered context products
**Is your feature request related to a problem? Please describe.**
I'm frustrated with the way the registered / non-registered product files are passed to validate when validate is integrated as part of another application.

**Describe the solution you'd like**
To have a public method in ValidateLauncher which takes the URL to the registered / non registered product file as argument and includes them in the validation process. A method like: 

- `public void setContextProducts(URL registeredContextProductsFile, URL nonRegisteredContextProductsFile)` 

or a set of methods like:

- `public void addRegisteredContextProducts(URL registeredContextProductsFile)`
- `public void addNonRegisteredContextProducts(URL nonRegisteredContextProductsFile)`

**Additional context**
BepiColombo has an application to pack pds4 products (pds4-packager) and it uses validate as its main validation step. I have integrated validate using `gov.nasa.pds.validate.ValidateLauncher` as entry point as it provides public methods for almost the same features the command line exposes.
To correctly validate BepiColombo products I need validate to use pds' context products file (which they are available in validate.jar - util/registered_context_products.json) and esa not-yet-registered-in-pds context products file (which they are available in other class path location).

The main problem I have at this point is that `ValidateLauncher` does not provide a public method to load nonRegisteredContextProducts file so I have to abuse `ValidateLauncher` to have them loaded. 

Moreover, to load registered / non registered context products, the corresponding files should be available in the directory pointed by the `resources.home` property. For me it would be more convenient if `ValidateLauncher` would load them form a couple of URL locations (like it does with the catalog file).

I use the following code to set up validate:

```
public Stream<ValidationMessage> apply(Pds4Label label) {

		try {
			// Copy context product to java.io.tmpdir directory and set resource.home property as expected by validate
			
			final File registeredContextProductsFile = new File(System.getProperty("java.io.tmpdir") + File.separator + "registered_context_products.json");
			final File localContextProductsFile = new File(System.getProperty("java.io.tmpdir") + File.separator + "local_context_products.json");
			
			FileUtils.copyURLToFile(registeredContextProductsUrl, registeredContextProductsFile);
			FileUtils.copyURLToFile(localContextProductsUrl, localContextProductsFile);
			System.setProperty("resources.home", System.getProperty("java.io.tmpdir"));
			
			// Configure validate
			
			final ValidateLauncher launcher = new ValidateLauncher();
			
			launcher.setUpdateRegisteredProducts(true);
			launcher.setNonRegisteredProducts(true);
			
			launcher.setCatalogs(Arrays.asList(catalogUrl.toExternalForm()));
			launcher.setSkipProductValidation(!checkData);
			launcher.setForce(true);			
			launcher.setMaxErrors(Long.MAX_VALUE);
			launcher.setReportStyle("json");
			launcher.setTargets(Arrays.asList(label.getLabelLocation().toExternalForm()));
			launcher.setupReport();
			
			// Send report to a string writer to retrieve later
				
			final Report report = launcher.getReport();
			final StringWriter writer = new StringWriter();
				
			report.setOutput(writer);
			report.printHeader();
			
			// Work-around to register non registered products file
			
			final Field nonRegisteredProductsFileField = ValidateLauncher.class.getDeclaredField("nonRegisteredProductsFile");
			
			nonRegisteredProductsFileField.setAccessible(true);
			nonRegisteredProductsFileField.set(launcher, localContextProductsFile);
			
			// Work-around to avoid context validation to raise a NullPointerException 
			
			final Method setRegisteredProducts = ValidateLauncher.class.getDeclaredMethod("setRegisteredProducts");
			
			setRegisteredProducts.setAccessible(true);
			setRegisteredProducts.invoke(launcher);

			// Do validation
				
			launcher.doValidation(new HashMap<URL, String>());
				
			report.printFooter();
				
			writer.close();
									
			// Parse JSON reply
				
			final ValidationReport validationReport = buildValidationReport(writer);
			
			// Return validation messages
				
			return validationReport.getMessages().stream()
					.filter(this::filterNonRegisteredProductWarning)
					.filter(this::filterInvalidGroupLengthErrors);
		}
		catch(IOException e) {
			throw new UncheckedIOException(e);
		}
		catch(ReflectiveOperationException e) {
			throw new IllegalStateException(e);
		}
		catch(Exception e) {
			throw new RuntimeException(e);
		}
	}
```

The requested change would:

1. Remove the need to copy resources that are already available in the class path and can be loaded directly.
1. Remove the need to abuse `ValidateLauncher` code calling private methods and fields through reflection. 

I would be also happy to re-do the integration code if there is a recommended way to integrate validate.

---

## Applicable requirements
🦄 #308 

## Engineering Details
Will add getter/setter classes to the `ValidateLauncher` class.

**Note for I&T Team:** I&T will be completed by customer as they can easily drop this into their pipeline to verify it meets their requirement needs

jordanpadams:
Resolves #241

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
+++ b/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
@@ -1034,7 +1034,7 @@ public class ValidateLauncher {
         List<ValidationProblem> pList = new ArrayList<ValidationProblem>();
 
         try {
-            JsonObject json = gson.fromJson(new FileReader(registeredProductsFile), JsonObject.class);
+            JsonObject json = gson.fromJson(new FileReader(this.registeredProductsFile), JsonObject.class);
             JsonArray array = json.get("Product_Context").getAsJsonArray();
 
             for (JsonElement jsonElm : array) {
@@ -1067,21 +1067,21 @@ public class ValidateLauncher {
             e.printStackTrace();
             System.out.println(e.getMessage()
                     + "\nInvalid JSON File: Verify format and values match that in RegisteredProducts File JSON file: "
-                    + registeredProductsFile);
+                    + this.registeredProductsFile);
             ValidationProblem pW = new ValidationProblem(
                     new ProblemDefinition(ExceptionType.ERROR, ProblemType.INTERNAL_ERROR,
                             e.getMessage()
                                     + "\nInvalid Registered Context Product JSON File: Verify format and values match that in RegisteredProducts File JSON file: "
-                                    + registeredProductsFile),
+                                    + this.registeredProductsFile),
                     url);
             pList.add(pW);
         }
         
-        if (nonRegisteredProducts) {
+        if (this.nonRegisteredProducts) {
             
             try {
                 gson = new Gson();
-                JsonObject jsonN = gson.fromJson(new FileReader(nonRegisteredProductsFile), JsonObject.class);
+                JsonObject jsonN = gson.fromJson(new FileReader(this.nonRegisteredProductsFile), JsonObject.class);
                 JsonArray arrayN = jsonN.get("Product_Context").getAsJsonArray();
                 ValidationProblem pW = new ValidationProblem(
                         new ProblemDefinition(ExceptionType.WARNING, ProblemType.NON_REGISTERED_PRODUCT,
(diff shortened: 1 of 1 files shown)


src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
```diff
@@ -1135,7 +1135,23 @@ public void setNonRegisteredProducts(boolean nonRegisteredProducts) {
         this.nonRegisteredProducts = nonRegisteredProducts;
     }
     
-    /**
+    public File getRegisteredProductsFile() {
+		return registeredProductsFile;
+	}
+
+	public void setRegisteredProductsFile(File registeredProductsFile) {
```
tloubrieu-jpl, line 1142: indentation ?

tloubrieu-jpl (approved):
Good !

PR #527, opened 2022-07-13
Set context products object later in execution

### LINKED ISSUE #526: Set registered context products in doValidation method
Hi @jordanpadams I have checked the latest snapshot and I would need a modification for our integration to work. 

From our integration code, calling doValidate() on gov.nasa.pds.validate.ValidateLauncher after calling setNonRegisteredProductsFile() and setRegisteredProductsFile() raises a NullPointerException. 

I have been debugging the code and I realised that on line 173  gov.nasa.pds.tools.validate.rule.pds4.ContextProductReferenceValidationRule.checkContextReferences() when checkContextReferences() retrieves the registered products from the context it gets a null object, causing a NullPointerException on line 220.

It seems that although the register / non registered context products have been set, they have not been parsed as the method used for parsing and put them in the context- ValidateLauncher.setRegisteredProducts() - is called in ValidateLauncher.processMain(String[] args) instead of in ValidateLaucher.doValidation() which is used for the integration.

I have tried to change the integration code to use ValidateLaucher.processMain(String args[]) instead of ValidateLauncher.doValidation() to work-around this error but that is not possible, mainly because processMain does it own report set-up and the integration code needs to customise the report before it writes anything.

To get the integration code working I would need that either ValidateLauncher.setRegisteredProducts() becomes public instead of private [so I can call it from the integration code before calling ValidateLauncher.doValidate()] or that the call to ValidateLauncher.setRegisteredProducts() in ValidateLauncher.processMain(String[] args) - line 1622 - gets moved inside ValidateLauncher.doValidate(). Could it be possible to include one of those modifications in the snapshot ?

Many thanks for your support.

_Originally posted by @fran-vallejo in https://github.com/NASA-PDS/validate/issues/241#issuecomment-1181788122_

## Related Requirements
:unicorn: #241

jordanpadams:
Executing `setRegisteredProducts` in `processMain` doesn't make sense for someone trying to call this class through the API.

Solution is the execution in the `doValidation` method, which encapsulates a lot of setup prior to executing validation.

Alternative would be to make `setRegisteredProducts` public.

Resolves #526

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
+++ b/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
@@ -1274,6 +1274,10 @@ public class ValidateLauncher {
     public boolean doValidation(Map<URL, String> checksumManifest) throws Exception {
         boolean success = true;
         long t0 = System.currentTimeMillis();
+
+        // Set the registered context products prior to looping through the targets
+        setRegisteredProducts();
+        
         // Initialize the Factory Class
         List<DocumentValidator> docValidators = new ArrayList<DocumentValidator>();
         factory = ValidatorFactory.getInstance();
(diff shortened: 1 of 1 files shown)


tloubrieu-jpl (approved):
That looks good to me

PR #571, opened 2022-12-05
validate-bundle needs quotes around $REPORT_DIR

VoyagerPWS:
validate-bundle needs quotes around $REPORT_DIR on line 105

When $REPORT_DIR has no value, the comparison is invalid without quotes.


jordanpadams (approved):

jordanpadams:
thanks @VoyagerPWS!

Issue #435, opened 2021-10-28
Array Content Validator is not accepting values at the min/max due to false precision


jstone-psi:
## 🐛 Describe the bug
The array content validator is flagging values as below the minimum or above the maximum that should be accepted. This seems to happen because single precision binary values are being converted to double precision, which is introducing false precision. 

If the value comparison methods were made using the ArrayContentValidator.compare method, which corrects for precision differences, this would likely resolve the issue.  This is how the scaled value comparisons already work.

## 📜 To Reproduce
Run the validator on the attached label and image:
```
validate flat_w.xml
```

This will yield the following error:

```
ERROR  [error.array.value_out_of_min_max_range]   array 1, location (1, 626): Value is less than the minimum value in the label (min=-0.25048923, got=-0.25048923).
```

## 🕵️ Expected behavior
Since the values are apparently equal, this should not be flagged as an error.

## 📚 Version of Software Used
validate 2.1.0

## 🩺 Test Data / Additional context
[flat_w.zip](https://github.com/NASA-PDS/validate/files/7437701/flat_w.zip)

## 🖥 System Info
 - OS: MacOS Big Sur
 - Java: openjdk 17

---

qchaupds:
As @jstone-psi had pointed out, there may be different in precision of the two values being compared.


```
{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 118 % vi src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java


  private void checkObjectStats(Number value, ElementArray elementArray,
      ObjectStatistics objectStats, ArrayLocation location) {
    if (objectStats.getMinimum() != null) {
      if (value.doubleValue() < objectStats.getMinimum()) {
        String errorMessage = this.tableNameReportStr + " Value is less than the minimum value in the label (min=" + objectStats.getMinimum().toString();
        LOG.debug("checkObjectStats:value.doubleValue() {}",value.doubleValue());
        LOG.debug("checkObjectStats:objectStats.getMinimum(),type(objectStats.getMinimum()) {},{}",objectStats.getMinimum(),objectStats.getMinimum().getClass().getSimpleName());
        LOG.error(errorMessage);
System.exit(0);
        addArrayProblem(ExceptionType.ERROR,
            ProblemType.ARRAY_VALUE_OUT_OF_MIN_MAX_RANGE,
            this.tableNameReportStr + " Value is less than the minimum value in the label (min="
[...]

PR #577, opened 2022-12-13
Update compare method to use BigDecimal.scale

### LINKED ISSUE #529: ERROR  [error.array.value_out_of_min_max_range] evaluation is not correct
When validating a pds4.folder of image type arrays, several errors were generated during the evaluation of the min/max array values.  Some were correct, others were not.

ERROR  [error.array.value_out_of_min_max_range]   array 1, location (349, 345): Value is greater than the maximum value in the label (max=0.0014194, got=0.0014194006).

This is technically correct, and we can fix it by increasing the significant digits in the data product.

 ERROR  [error.array.value_out_of_min_max_range]   array 1, location (386, 185): Value is greater than the maximum value in the label (max=0.0702686, got=0.07026859).

This is not correct.

My assumption is that validate is looking at the number of significant digits, and is giving incorrect results when the number of digits is different.  

I can attach test case data products if needed.

## Engineering Details
Tightly coupled with #514 and #434

see #574 for additional info

jordanpadams:
Fix 1
-----
Original compare method had some confusions between precision and scale. We should be using scale for handling doubles with many significant digits.

> The precision is the number of digits in the unscaled value. For instance, for the number 123.45, the precision returned is 5.

> If zero or positive, the scale is the number of digits to the right of the decimal point. If negative, the unscaled value of the number is multiplied by ten to the power of the negation of the scale. For example, a scale of -3 means the unscaled value is multiplied by 1000.

Additionally, we need to be able to handle scale appropriately. Previous functionality would decrease the scale to the least between the min/max value and the actual value. This obviously doesn't make sense because when we do this, comparisons like this will be invalid:

actual value: 0.0014194006
scaled actual value: 0.0014194
max value from label: 0.0014194

as you can see, if we compare the max value from the label to the scaled actual value, they are equal. but in reality they are not. this should raise an error.

Fix 2
-----
that being said, we can't always trust the full scaled values that are calculated because doubles can introduce false precision. if we just change the algorithm to use whichever is the largest scale (aka number of decimal places) we were taking values that sometimes introduced false precision. thanks for our regression test in #435, we were able to see that the solution is not as simple as increasing the scale. so after some reading, it looks like using the BigDecimal(String) initializer as much as possible, and avoiding calculations/conversions to double in memory is the best way to avoid false precision. there are still some possible issues with false precision in the code, but this should resolve most of them.

resolves #529

References:
* https://blogs.oracle.com/javamagazine/post/four-common-pitfalls-of-the-bigdecimal-class-and-how-to-avoid-them
* https://stackoverflow.com/questions/35435691/bigdecimal-precision-and-scale

## ⚙️ Test Data and/or Report
https://github.com/NASA-PDS/validate/actions/runs/3688227933


tloubrieu-jpl (approved):
That looks good to me.

Issue #556, opened 2022-11-16
As a user, I want to be able to use validate from a docker container


jordanpadams:
## 🧑‍🔬 User Persona(s)

Data Provider, Node Personnel

## 💪 Motivation
...so that I do not need to install Java in order to run the app.

## ⚖️ Acceptance Criteria
**Given** 
**When I perform** 
**Then I expect**

ramesh-maddegoda:
Created the pull request https://github.com/NASA-PDS/validate/pull/580

ramesh-maddegoda:
Incorporated pull request review comments at https://github.com/NASA-PDS/validate/pulls

jordanpadams:
done per #580

miguelp1986:
Testrail: https://cae-testrail.jpl.nasa.gov/testrail/index.php?/cases/view/1864370

PR #580, opened 2022-12-22
Dockerize validate 556

ramesh-maddegoda:
## 🗒️ Summary
This pull request is created to Dockerize the Validate tool. The docker image of Validate Tool provides an easy way to deploy the Validate Tool, without having to  worry about the Java/JDK versions installed and operating systems supported. Also, this helps to  easily use the Validate Tool in other execution environments such as docker compose and Amazon ECS.

## ♻️ Related Issues
NASA-PDS/validate#556

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -34,8 +34,8 @@ FROM openjdk:15-slim
 ARG validate_version=3.0.3
 
 # Install curl (to download the validate binary)
-RUN apt-get update -y
-RUN apt-get install curl -y
+RUN apt-get update -y \
+    && apt-get install -y curl
 
 ENV VALIDATE_IMAGE_PATH=https://github.com/NASA-PDS/validate/releases/download/v${validate_version}/validate-${validate_version}-bin.tar.gz
 
@@ -43,9 +43,10 @@ ENV VALIDATE_IMAGE_PATH=https://github.com/NASA-PDS/validate/releases/download/v
 ADD $VALIDATE_IMAGE_PATH /tmp/validate-bin.tar.gz
 RUN  mkdir /opt/validate  \
      && tar xzf /tmp/validate-bin.tar.gz  -C /opt/validate --strip-components 1 \
-     && rm -f /tmp/validate-bin.tar.gz
+     && rm -f /tmp/validate-bin.tar.gz \
+     && apt-get remove --purge \
+     && rm -rf /var/lib/apt/lists/*
+
 ENV PATH="$PATH:/opt/validate/bin"
 
-# Pass default arguments to validate tool
-#CMD validate $DEFAULT_ARGS
 ENTRYPOINT ["validate"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,50 @@
+# Copyright 2022, California Institute of Technology ("Caltech").
+# U.S. Government sponsorship acknowledged.
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice,
+# this list of conditions and the following disclaimer.
+# * Redistributions must reproduce the above copyright notice, this list of
+# conditions and the following disclaimer in the documentation and/or other
+# materials provided with the distribution.
+# * Neither the name of Caltech nor its operating division, the Jet Propulsion
+# Laboratory, nor the names of its contributors may be used to endorse or
+# promote products derived from this software without specific prior written
+# permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+FROM openjdk:15-slim
+
+# Set following argument with the required validate tool version
+ARG validate_version=3.0.3
+
+# Install curl (to download the validate binary)
+RUN apt-get update -y
```
nutjob4life: Combine multiple `apt-get` statements with `&&` in order to reduce the number of layers/image size.

docker/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN apt-get install curl -y
+
+ENV VALIDATE_IMAGE_PATH=https://github.com/NASA-PDS/validate/releases/download/v${validate_version}/validate-${validate_version}-bin.tar.gz
+
+# Install Validate tool
+ADD $VALIDATE_IMAGE_PATH /tmp/validate-bin.tar.gz
+RUN  mkdir /opt/validate  \
+     && tar xzf /tmp/validate-bin.tar.gz  -C /opt/validate --strip-components 1 \
+     && rm -f /tmp/validate-bin.tar.gz
```
nutjob4life: Consider adding `apt-get remove --purge` and/or `rm -rf /var/lib/apt/lists/*` as a final clean-up step to make the image smaller.

docker/README.md
```diff
@@ -0,0 +1,130 @@
...
+# Set following argument with the required validate tool version
+ARG validate_version=3.0.3
+```
+
+3. Open a terminal and change the current working directory to `validate/docker`.
+
+4. Build the docker image as follows.
+
+```
+    docker image build -t nasapds/validate .
```
nutjob4life: In documentation, I really like to see the full arguments like `--tag` instead of `-t` 😇

docker/README.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+5. Test the `nasapds/validate` docker image as follows.
+
+```
+    docker container run nasapds/validate --help
```
nutjob4life: I thing there should be `--rm` after `run` otherwise the user is going to be left with a lot of stale containers and loss of disk space

docker/README.md
```diff
@@ -0,0 +1,130 @@
...
+[Operation documentation](https://nasa-pds.github.io/validate/operate/index.html) of the Validate tool.
+
+```shell
+% validate $HOME/pds/bundle -M checksum-manifest.txt -r validate-report.txt -R pds4.bundle
+```
+
+However, when using the dockerized Validate Tool, above command can be updated as follows.
+
+```shell
+docker container run --name validate \
```
nutjob4life: Can drop `--name validate` since this container is going to be `--rm`oved 😀

docker/README.md
```diff
@@ -0,0 +1,130 @@
...
+                 --rm \
+                 --volume "${HOME}/pds/bundle":/tmp/pds \
+                 --volume "${HOME}/pds/bundle/checksum-manifest.txt":/tmp/pds/checksum-manifest.txt \
+                 --volume "${HOME}/validate-report":/tmp/validate-report \
+                 nasapds/validate /tmp/pds -M /tmp/pds/checksum-manifest.txt -r /tmp/validate-report/validate-report.txt -R pds4.bundle
+```
```
nutjob4life, line 74: Otherwise this is beautiful! ✨

nutjob4life (changes requested):
Only minor things

tloubrieu-jpl (approved):
That works for me, besides @nutjob4life comments. 

As a side thought I am wondering how the volumes are going to be actually mounted, like the checksum file should be on the same volume as the bundle itself (the archive, read-only), and the report will be written in user space. That will be for later and will depend on how validate will be upgrade to work on s3 and/or integrated in nucleus.

ramesh-maddegoda:
> That works for me, besides @nutjob4life comments.
> 
> As a side thought I am wondering how the volumes are going to be actually mounted, like the checksum file should be on the same volume as the bundle itself (the archive, read-only), and the report will be written in user space. That will be for later and will depend on how validate will be upgrade to work on s3 and/or integrated in nucleus.

@tloubrieu-jpl, if we can safely assume that the checksum file should be on the same volume as the bundle itself, then we can reduce the number of volume parameters.

jordanpadams:
> > That works for me, besides @nutjob4life comments.
> > As a side thought I am wondering how the volumes are going to be actually mounted, like the checksum file should be on the same volume as the bundle itself (the archive, read-only), and the report will be written in user space. That will be for later and will depend on how validate will be upgrade to work on s3 and/or integrated in nucleus.
> 
> @tloubrieu-jpl, if we can safely assume that the checksum file should be on the same volume as the bundle itself, then we can reduce the number of volume parameters.

@ramesh-maddegoda we cannot really assume anything about where the checksum file resides. additionally, that file is optional to begin with.

jordanpadams:
@ramesh-maddegoda @nutjob4life just to confirm, we are a go for merge here?

ramesh-maddegoda:
> @ramesh-maddegoda @nutjob4life just to confirm, we are a go for merge here?

@jordanpadams , not yet. I am making few code improvements

docker/Dockerfile
ramesh-maddegoda: done

docker/Dockerfile
ramesh-maddegoda: done

docker/README.md
ramesh-maddegoda: done

docker/README.md
ramesh-maddegoda: done

ramesh-maddegoda:
> @ramesh-maddegoda @nutjob4life just to confirm, we are a go for merge here?

@jordanpadams and @nutjob4life , I incorporated all the review comments! Thanks!

docker/run.sh
```diff
@@ -0,0 +1,64 @@
...
+CHECKSUM_MANIFEST_FILE_NAME=urn-nasa-pds-insight_rad.md5
+
+# Validate report directory path
+VALIDATE_REPORT_PATH=${HOME}
+
+# Validate report file name
+VALIDATE_REPORT_FILE_NAME=validate-report.txt
+
+# Default arguments provided as an array. This can be updated based on the requirements.
+DEFAULT_ARGS=(/tmp/pds -M /tmp/pds/${CHECKSUM_MANIFEST_FILE_NAME} -r /tmp/validate-report/${VALIDATE_REPORT_FILE_NAME} -R pds4.bundle)
```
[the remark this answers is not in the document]
ramesh-maddegoda: Updated the code to use bash shell as suggested in https://github.com/koalaman/shellcheck/wiki/SC3030

docker/run.sh
```diff
@@ -0,0 +1,64 @@
...
+
+echo "Executing: validate ${DEFAULT_ARGS[*]}"
```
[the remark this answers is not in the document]
ramesh-maddegoda: Updated the code to use bash shell as suggested in https://github.com/koalaman/shellcheck/wiki/SC3054

jordanpadams (approved):

PR #588, opened 2023-01-26
issue #514: table not finding NaNs and Infs

### LINKED ISSUE #514: Validate does not catch NaNs in Binary Tables
I found Nan in Binary_Table. The validation tool didn't give me an error. I was looking at the data and I thought the nan. Is the validation tool not supposed to return an error? Also I can't upload file so I will see how I can send the example file to you.

Tightly coupled with #529 and #434

al-niessner:
## 🗒️ Summary
pds4-jparser uses the Java built-in Double.toString() to convert the values for tables. It means that the code is dependent upon how the JVM implements that conversion. Seems INF and NAN are no longer vogue and it now returns Infinity and NaN. Updated the available checks to include INFINITY and then convert the Double.toString().toUpperCase() for the comparison only for running with any JVM version.

## ⚙️ Test Data and/or Report
Added test to auto checks so look for green check below.

## ♻️ Related Issues
resolves #514


nutjob4life (approved):
Thanks ~~Sun~~ ~~Java~~ Oracle for changing things on us. 👍

nutjob4life (approved):
Looks even better. Is this all of it? If so: 👍

Issue #325, opened 2021-04-09
Validate Incorrectly Throws Error When Embedded Field_Character Contains <CR><LF>


radiosci:
## 🐛 Describe the bug
I have Table_Character files with long records.  Carriage-Return Line-Feed pairs have been embedded within the records to facilitate display; but the original (long) record length has been retained in the label.  Validate finds the embedded delimiter pairs, claims my records are too short, and throws errors.  
A prohibition on embedded Line-Feed and Carriage-Return characters was approved for Field_Delimited in CCB-181; a broader prohibition on non-printing control characters in Field_Character was proposed in CCB-196, but it was rejected by DDWG.  That presumably means Carriage-Return Line-Feed pairs may appear in embedded Field_Character so long as the records are correctly delimited by Carriage-Return Line-Feed.
Validate should return errors if character records are incorrectly delimited; it should not return errors if delimiter pairs are found internal to records.

## 📜 To Reproduce
1. Run validate on the attached crs009x.{tab, xml} product
2. See error report (crs009x_report.txt)

## 🕵️ Expected behavior
crs009x.{tab, xml} should pass.

## 📚 Version of Software Used
validate 1.24.0

## 🩺 Test Data / Additional context
[crs009x.tab.gz](https://github.com/NASA-PDS/validate/files/6288046/crs009x.tab.gz)
[crs009x.xml.gz](https://github.com/NASA-PDS/validate/files/6288047/crs009x.xml.gz)
[crs009x_report.txt](https://github.com/NASA-PDS/validate/files/6288041/crs009x_report.txt)

##  🏞Screenshots
N/A

## 🖥 System Info
 - OS: Mac OSX 11.1
 - Browser: Safari 14.0.2
 - Version

---

** :unicorn: Applicable requirements**

LevN0:
Without commenting on what the validate tool should do,

>I have Table_Character files with long records. Carriage-Return Line-Feed pairs have been embedded within the records to facilitate display; but the original (long) record length has been retained in the label

This is a good way to make plain-ASCII data much more difficult to read for any non-specialized tool, for seemingly no benefit. Tools following standard will read it, but general tools shipping with the programming languages will choke. 

Embedded linebreaks facilitate display only if everyone has an agreed upon window size, otherwise they make the display worse.

radiosci:
[...]

Issue #379, opened 2021-07-27
FileService:printStackTraceToFile:ERROR when validating a product with overlapping fields


fran-vallejo:
## 🐛 Describe the bug
Validating a product that contains a table with overlapping fields results in an error when validate tries to create a stack trace report file.

## 📜 To Reproduce
Run validate (either version 2.0.7 or 2.1.0-SNAPSHOT) using **mix_cal_hk_fpac_report_20181204** as target.

```
BepiVirtualBox:~/Software$ validate-2.0.7/bin/validate --target /home/fvallejo/Documents/mix_cal_hk_fpac_report_20181204.xml
FileService:printStackTraceToFile:ERROR: This function can only write to file that ends with .txt or .log

```

## 🕵️ Expected behavior
Validate runs and reports the product is not valid because it defines overlapping fields.

## 📚 Version of Software Used
validate 2.0.7 and validate 2.1.0-SNAPSHOT [repo commit 2cba1fd32441626e30669d6cd4fa1bc6f8088d5f]

## 🩺 Test Data / Additional context
Product [mix_cal_hk_fpac_report_20181204.zip](https://github.com/NASA-PDS/validate/files/6884980/mix_cal_hk_fpac_report_20181204.zip) attached. Fields 42 and 43 in this product overlap.

Please note that validate will report missing external references as well as the product contains references to elements not yet in PDS.

## 🖥 System Info
 - OS: Ubuntu 20.04
 - Java: OpenJDK 64-Bit Server VM (build 11.0.11+9-Ubuntu-0ubuntu2.20.04, mixed mode, sharing)

---

jstone-psi:
This happens to me, also. I managed to track the problem to `TableDataContentValidationRule.java`, where the stack trace logger is trying to write to the data file instead of a log file based on the data file name.

@jordanpadams I think this can be fixed by changing

```
FileService.printStackTraceToFile(dataFile.getPath(),ex);
```

to

```          
FileService.printStackTraceToFile(dataFile.getPath() + ".log",ex);
```

in three places in `TableDataContentValidationRule.java`.

qchaupds:
@jstone-psi @jordanpadams You are right, the call to printStackTraceToFile() should pass in either "null", or appends ".log".  The default name will be used if "null" is passed in.
With some changes to our Java code, we have success:

[...]

PR #594, opened 2023-02-08
issue #593: fix regression introduced to #325 integration tests

### LINKED ISSUE #593: Regression in validate no longer enabling CRLF to be embedded within a Table_Character record
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

Per [updates for](https://github.com/NASA-PDS/validate/pull/589) #519, added a regression for a failed test related to #325.

It appears that somewhere in the updates a regression was added where we are reading in `Table_Character` like `Table_Delimited`. So this special case (among probably others) are no longer handled correctly.

### 🕵️ Expected behavior

Validate successfully

### 📜 To Reproduce

Previously when running `mvn test`:

```
[ERROR] Execute validate command for tests below. #87  Time elapsed: 0.755 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 0 error messages expected. {"title":"PDS Validation Tool Report","configuration":{"version":"3.2.0-SNAPSHOT","date":"2023-01-30T22:02:41Z"},"parameters":{"targets":"[file:/Users/jpadams/proj/pds/pdsen/workspace/validate/src/test/resources/github325/crs009x.xml]","ruleType":"pds4.label","severityLevel":"WARNING","recurseDirectories":"true","fileFiltersUsed":"[*.xml, *.XML]","dataContentValidation":"on","productLevelValidation":"on","maxErrors":"100000","registeredContextsFile":"/Users/jpadams/proj/pds/pdsen/workspace/validate/src/main/resources/util/registered_context_products.json"},"productLevelValidationResults":[{"status":"FAIL","label":"file:/Users/jpadams/proj/pds/pdsen/workspace/validate/src/test/resources/github325/crs009x.xml","messages":[],"fragments":[],"dataContents":[{"dataFile":"file:/Users/jpadams/proj/pds/pdsen/workspace/validate/src/test/resources/github325/crs009x.tab","messages":[{"severity":"ERROR","type":"error.table.record_length_mismatch","table":1,"record":1,"message":"Delimiter is not at the end of the record. Record read using delimiter is 82 bytes long while record is defined to be 246 bytes."},{"severity":"ERROR","type":"error.table.record_length_mismatch","table":1,"record":2,"message":"Delimiter is not at the end of the record. Record read using delimiter is 82 bytes long while record is defined to be 246 bytes."}]}]}],"summary":{"totalProducts":1,"totalErrors":2,"totalWarnings":0,"messageTypes":[{"messageType":"error.table.record_length_mismatch","total":2}]}} ==> expected: <2> but was: <0>
```

Now, executing validate against that file in the repo, instead of the test executing successfully, we get 2 errors:
```
$ validate -t src/test/resources/github325/crs009x.xml
```

### 🖥 Environment Info

- Version of this software [e.g. vX.Y.Z]
- Operating System: [e.g. MacOSX with Docker Desktop vX.Y]
...

_No response_

See repo path in execution above

### 🦄 Related requirements

_No response_

### ⚙️ Engineering Details

_No response_

al-niessner:
## 🗒️ Summary
Many part fix. In this repository, change the table validator to push character tables to the correct test. However, it requires the changes waiting in NASA-PDS/pds4-jparser#86.

## ⚙️ Test Data and/or Report
Test github325 is enabled and should pass as well as all other unit tests.

## ♻️ Related Issues
fixes #593

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
@@ -150,7 +150,8 @@ public class TableValidator implements DataObjectValidator {
       if (this.tableAdapter instanceof TableBinaryAdapter) {
         this.validateTableBinaryContent(fieldValueValidator, record, spotCheckData,
             keepQuotationsFlag);
-      } else if (!this.isTableLineOriented() && !this.getCheckInbetweenFields()) {
+      } else if (this.tableAdapter instanceof TableDelimitedAdapter &&
+    		  !this.isTableLineOriented() && !this.getCheckInbetweenFields()) {
         // Determine if we should proceed with calling validateTableDelimited()
         // function.
         // Note that the function validateTableDelimited() can only be applied if the
@@ -444,7 +445,7 @@ public class TableValidator implements DataObjectValidator {
     // reduced the function size.
     TableCharacterUtil tableCharacterUtil = null;
     boolean manuallyParseRecord = false;
-    String line = this.currentTableReader.readNextLine();
+    String line = this.currentTableReader.readNextFixedLine();
     long lineNumber = 0;
     int dataObjectIndex = this.tableObject.getDataObjectLocation().getDataObject();
 
@@ -609,7 +610,7 @@ public class TableValidator implements DataObjectValidator {
         break;
       }
 
-      line = this.currentTableReader.readNextLine();
+      line = this.currentTableReader.readNextFixedLine();
       if (line != null) {
         LOG.debug("recordNumber: {}, line.length: {}, record: [{}]", lineNumber, line.length(),
             line);
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
@@ -19,6 +19,7 @@ import java.net.URL;
 import java.nio.BufferUnderflowException;
 import java.util.ArrayList;
 import java.util.Map;
+import org.checkerframework.common.returnsreceiver.qual.This;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import com.opencsv.exceptions.CsvValidationException;
@@ -445,7 +446,7 @@ public class TableValidator implements DataObjectValidator {
     // reduced the function size.
     TableCharacterUtil tableCharacterUtil = null;
     boolean manuallyParseRecord = false;
-    String line = this.currentTableReader.readNextFixedLine();
+    String line = this.tableAdapter instanceof TableDelimitedAdapter ? this.currentTableReader.readNextLine() : this.currentTableReader.readNextFixedLine();
     long lineNumber = 0;
     int dataObjectIndex = this.tableObject.getDataObjectLocation().getDataObject();
 
@@ -610,7 +611,7 @@ public class TableValidator implements DataObjectValidator {
         break;
       }
 
-      line = this.currentTableReader.readNextFixedLine();
+      line = this.tableAdapter instanceof TableDelimitedAdapter ? this.currentTableReader.readNextLine() : this.currentTableReader.readNextFixedLine();
       if (line != null) {
         LOG.debug("recordNumber: {}, line.length: {}, record: [{}]", lineNumber, line.length(),
             line);
--- a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
@@ -449,7 +449,7 @@ public class TableValidator implements DataObjectValidator {
     // reduced the function size.
     TableCharacterUtil tableCharacterUtil = null;
     boolean manuallyParseRecord = false;
-    String line = this.tableAdapter instanceof TableDelimitedAdapter ? this.currentTableReader.readNextLine() : this.currentTableReader.readNextFixedLine();
+    String line;
     long lineNumber = 0;
     int dataObjectIndex = this.tableObject.getDataObjectLocation().getDataObject();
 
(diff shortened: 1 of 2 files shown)


al-niessner:
@jordanpadams @nutjob4life @tloubrieu-jpl 

Can we update all pds4-jarser again and then let these unit tests run. I want to see if my changes break something else. Looks good on my laptop but it does not always translate to jenkins.

jordanpadams:
@al-niessner merged pds4-jparser and re-ran the integration tests

jordanpadams:
@al-niessner also can you update the PR description to use the [GitHub keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) to close the ticket when we merge this (e.g. fixes #593). Just one less step for us to track down the ticket and close it out

jordanpadams:
@al-niessner looks like the tests failed with the latest pds4-jparser

jordanpadams:
https://github.com/NASA-PDS/validate/actions/runs/4129185892

al-niessner:
@jordanpadams 

Yes, I got the email that it failed but it was another test. Thought that might be the case because that is what happened earlier in previous ticket that broke 325. Will see why it breaks and keep going until it all turns green. Thanks.

Also, I know about the keywords but thought you did not want to use them for some reason. I can use them easy enough.

jordanpadams:
> Also, I know about the keywords but thought you did not want to use them for some reason. I can use them easy enough.

my bad. thanks!

al-niessner:
@jordanpadams @nutjob4life

Okay, I do not understand where the error is here (code, test, or concept). I will write it out and hopefully never post it because I figured it out along the way. The new table validation does a better job?? and this error now seems to be doing what it is supposed to do but not what it did. So what is right? I will say that it is because now reading based on a fixed length rather than delimiter and given the table is character that seems correct as well.

There is an error in the table that fields overlap (#379). In the old version, got this error and only this error:

```
      ERROR  [error.table.field_value_overlap]   data object 1, record 1, field 42: This field overlaps the next field. Current field ends at 347. Next field starts at 347 but should be at least at 348
```

However this is back when the character table was being processed as a delimited table. Either way of processing the table ends up here: 

https://github.com/NASA-PDS/validate/blob/db5bcd4f54a25380b8407e3016c3882c73ef6db8/src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java#L300

In the delimited case, it breaks out of row processing immediately, does not tell the user that it did not skip rows, and moves onward with that single error.

In the character case, it generates these errors:

```
      ERROR  [error.table.field_value_overlap]   data object 1, record 1, field 42: This field overlaps the next field. Current field ends at 347. Next field starts at 346 but should be at least at 348
      ERROR  [error.table.field_value_overlap]   data object 1, record 1, field 42: This field overlaps the next field. Current field ends at 347. Next field starts at 347 but should be at least at 348
      ERROR  [error.table.records_mismatch]   data object 1: Number of records read is not equal to the defined number of records in the label (expected 583, got 1).
```

The first error is the same, thank goodness since the code is identical. The second error is new for reasons that I cannot explain since the error is in the middle of the record (741 bytes) but maybe for the same reason as the third error and I have not discovered it yet: delimiter table and character table do not check all of the same stuff. The third error occurs back here:

https://github.com/NASA-PDS/validate/blob/db5bcd4f54a25380b8407e3016c3882c73ef6db8/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java#L649

There is no similar check anywhere in [validateTableDelimitedContent(FieldValueValidator, URL, int, boolean)](https://github.com/NASA-PDS/validate/blob/db5bcd4f54a25380b8407e3016c3882c73ef6db8/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java#L124-L201).

While the third error seems to point to updating the validate.features file, the second error implies not so much - mostly because I cannot explain it. When I look at [XML for test](https://github.com/NASA-PDS/validate/blob/db5bcd4f54a25380b8407e3016c3882c73ef6db8/src/test/resources/github379/mix_cal_hk_fpac_report_20181204.xml#L613-L636) the overlap is clear and the two messages from character table seems correct. Just not sure why the delimited table is not showing it since it appears to be happening the the shared code.

So, what should be done? The shortest and maybe near-term-correct is to update validate.features for the two new errors but what about the contradictions with delimited table processing? Ignore it? Or is time for the real fix to all use the same table checking but just have different readers (way more work with scary unknown ripples)?

jordanpadams:
> So, what should be done? The shortest and maybe near-term-correct is to update validate.features for the two new errors but what about the contradictions with delimited table processing? Ignore it? Or is time for the real fix to all use the same table checking but just have different readers (way more work with scary unknown ripples)?

@al-niessner I say we fix the test for now, and see if other issues arise... I agree that the second error may just occur because when the first error gets caught, the reader gets all thrown off and may just throw another? either way, I think we march forward and see if we break anything. once you update the validate.features I will merge this PR.

per the contradictions with delimited processing, are you saying there is no `RECORDS_MISMATCH` warning anywhere in the handling of delimited tables?

al-niessner:
@jordanpadams 

> per the contradictions with delimited processing, are you saying there is no `RECORDS_MISMATCH` warning anywhere in the handling of delimited tables?

Yes.

al-niessner:
@jordanpadams @nutjob4life 

All yours to review. Seems that cleared up all of the regression tests.

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
```diff
@@ -470,6 +469,8 @@ private void validateTableCharacterContent(FieldValueValidator fieldValueValidat
     ArrayList<Integer> lineLengthsArray = new ArrayList<>(0);
     ArrayList<Long> lineNumbersArray = new ArrayList<>(0);
 
+    line = tableIsFixedLength ? this.currentTableReader.readNextFixedLine() : this.currentTableReader.readNextLine();
```
nutjob4life, line 472: It feels like the class of `currentTableReader` should provide a single API `readNextLine` and it itself figures out if it needs to provide a fixed line or a next line rather than relying on clients of the object to do it.

But I know, I know, I sound like a broken record. This is just the way "validate" is architected! 😝

nutjob4life (approved):
LGTM 👍

src/main/java/gov/nasa/pds/tools/validate/rule/pds4/TableValidator.java
al-niessner: Actually, no. It is not the reader that thinks it is fixed length. It is the XML definition that thinks it should be then the test code tells the reader to do a fixed length read rather delimited based on 3 values in the XML - touch more messed up than you were giving it credit for.

Your points are correct: Yes, some XML adapter should be asked if it is fixed length. Or, yes, the reader should be more complete and use the XML to do the necessary read (jparser). Yes, nothing will change because "do not fix it if it is not broken".

jordanpadams: @nutjob4life @al-niessner if you have a proposal for an architectural improvement to the software, feel free to create a ticket for it, and we can see if it becomes a higher priority in the future. I am all good with "do not fix it if it is not broken", until the technical debt becomes too high for us to manage. @al-niessner let me know when you think we have hit that tipping point.

PR #618, opened 2023-03-30
issue 616: add unit test first

al-niessner:
## 🗒️ Summary
The actual fix is in the NASA-PDS/pds4-jparser#89 that uses table sizes rather than file sizes. This PR is required to add regression testing to validate but requires the fixes on j
parser PR.

## ⚙️ Test Data and/or Report
See unit testing

## ♻️ Related Issues
Does not close a ticket but does support regression testing


al-niessner:
@jordanpadams @nutjob4life @tloubrieu-jpl 

Needs the jparser PR then updated to use it before merging to main.

nutjob4life (approved):
👍

PR #566, opened 2022-12-01
Update open/close of object reader

jordanpadams:
Previous functionality opened and closed the reader within a recursive method. Moved thie open/close up a level in execution.

Resolves #564

## ⚙️ Test Data and/or Report
```
$ validate-3.1.0-SNAPSHOT/bin/validate -t ~/test/geo_issue_20221201/s_07171001_sim.xml

PDS Validate Tool Report

Configuration:
   Version                       3.1.0-SNAPSHOT
   Date                          2022-12-01T17:26:25Z

Parameters:
   Targets                       [file:/Users/jpadams/test/geo_issue_20221201/s_07171001_sim.xml]
   Severity Level                WARNING
   Recurse Directories           true
   File Filters Used             [*.xml, *.XML]
   Data Content Validation       on
   Product Level Validation      on
   Max Errors                    100000
   Registered Contexts File      /Users/jpadams/proj/pds/pdsen/workspace/validate/validate-3.1.0-SNAPSHOT/resources/registered_context_products.json



Product Level Validation Results

  PASS: file:/Users/jpadams/test/geo_issue_20221201/s_07171001_sim.xml
        1 product validation(s) completed

Summary:

  0 error(s)
  0 warning(s)

  Product Validation Summary:
    1          product(s) passed
    0          product(s) failed
    0          product(s) skipped

  Referential Integrity Check Summary:
    0          check(s) passed
    0          check(s) failed
    0          check(s) skipped


End of Report
Completed execution in 9730 ms
```

Test data emailed to reviewers via LFT

PR #627, opened 2023-04-10
Provide enhanced exception logging for Array validation per #566

### LINKED ISSUE #567: Update ArrayObject exceptions to enable improved error messaging
> Asking the user to look at data type vs size is not going to go over well in the end. The checks in ArrayObject that check dimensions and index can return different (less generic) exceptions to produce a meaningful message appropriate the actual failure. If the reference to data type is data type size vs number of total bytes available then that should be simple to check as well given actual type is returned by ArrayObject.

@al-niessner totally agree. will create a separate ticket to refactor ArrayObject to provide better context to the user.

_Originally posted by @jordanpadams in https://github.com/NASA-PDS/validate/issues/566#issuecomment-1335669433_

al-niessner:
## 🗒️ Summary
Treat the various purposely thrown exceptions independently to improve the messaging. Still generates a similar error causing the same problem report but different messages. While not exhaustive, put more text in message to report the generic messages to fix the original throw to put the cause in the exception as a message.
 
## ⚙️ Test Data and/or Report
See unit tests of this PR

## ♻️ Related Issues
Closes #567

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java
@@ -252,6 +252,16 @@ public class ArrayContentValidator {
           LOG.warn("validatePosition:Unhandled dataType {}", dataType);
           break;
       }
+    } catch (ArrayIndexOutOfBoundsException aiobe) {
+      String message = "No message specified, but probably index out of range. Report this message to have it fixed (have original exception given a message";
+      if (aiobe.getMessage() != null)
+        message = aiobe.getMessage();
+      throw new IOException(message);
+    } catch (IllegalArgumentException iae) {
+      String message = "No message specified, but probably wrong number of dimensions. Report this message to have it fixed (have original exception given a message";
+      if (iae.getMessage() != null)
+        message = iae.getMessage();
+      throw new IOException(message);
     } catch (Exception ee) {
       String loc = Arrays.toString(location.getLocation());
       if (location.getLocation().length > 1) {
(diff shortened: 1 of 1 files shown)


src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java
```diff
@@ -252,6 +252,16 @@ private void validatePosition(Array array, ArrayObject arrayObject, ArrayLocatio
           LOG.warn("validatePosition:Unhandled dataType {}", dataType);
           break;
       }
+    } catch (ArrayIndexOutOfBoundsException aiobe) {
+      String message = "No message specified, but probably index out of range. Report this message to have it fixed (have original exception given a message";
```
nutjob4life: Extreme pedant here: closing `)` to match the `(` surrounding `have original exception given a message`.

src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java
```diff
@@ -252,6 +252,16 @@ private void validatePosition(Array array, ArrayObject arrayObject, ArrayLocatio
...
+      if (aiobe.getMessage() != null)
+        message = aiobe.getMessage();
+      throw new IOException(message);
+    } catch (IllegalArgumentException iae) {
+      String message = "No message specified, but probably wrong number of dimensions. Report this message to have it fixed (have original exception given a message";
```
nutjob4life: Ditto. (Extreme pedants are the worst! 😆)

nutjob4life (commented):
This is fine; don't worry about the `)` … I'm just being annoying 😝

nutjob4life (approved):

src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java
al-niessner: fixed

src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java
al-niessner: fixed

nutjob4life (approved):
This makes me so happy :)))))))))))))

Issue #534, opened 2022-08-19
B14.0 Content Validation Improvements: Additional Table Types, Additional File Areas


jordanpadams:
## 💡 Description

See sub-tasks for more details

jordanpadams:
📆 February status: Will be deferred in order to work on other higher priority Registry tasks. Per stakeholder feedback, will include in point build as needed.

al-niessner:
@jordanpadams All sub tasks are closed. I am closing this one but you can reopen it if should not be closed.

Issue #343, opened 2021-05-14
As a user I want to see the name of a table/array in errors, if one is specified


msbentley:
## Motivation

Errors like this:
```
[ERROR] e.b.p.v.Pds4ValidatorImpl: srn_cal_sc_strofio_hk-status_20200429_20200429.cdf - array:55 location:70 Value is not within the valid range of the data type 'IEEE754LSBDouble': NaN
```
can be difficult to track down in objects with many tables/arrays. Rather than counting them in the label, it would be great if the array or table name (if specified, given that it's optional) could be displayed, e.g.
```
[ERROR] e.b.p.v.Pds4ValidatorImpl: srn_cal_sc_strofio_hk-status_20200429_20200429.cdf - array:55 (HV_NEG_1_5KV_LEVEL) location:70 Value is not within the valid range of the data type 'IEEE754LSBDouble': NaN
```

## Acceptance Criteria
**Given** That a table or array has a `name` attribute
**When I perform** a validation check which fails on this table/array
**Then I expect** the name of the table/array to be displayed in the output

qchaupds:
@msbentley Help me to understand what the table or array name may look like?  
Better yet,  if you have an example label and data that you can share that may help me to see which code is affected when validate is ran against it.

qchaupds:
What I am hinting at is how does the label specify the table or array name?  Is it the "local_identifier" tag?

An example:

{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate/src/test/resources/github343/test_data 109 % view science_index_good.xml

     80     <Table_Character>
     81       <local_identifier>TABLE</local_identifier>
     82       <offset unit="byte">0</offset>
     83       <records>3</records>
     84       <description>Science Index Table</description>
     85       <record_delimiter>Carriage-Return Line-Feed</record_delimiter>
     86       <Record_Character>
     87         <fields>14</fields>

msbentley:
Hi @qchaupds - I meant it quite literally, the `name` attribute ;-) e.g.

https://pds.nasa.gov/datastandards/documents/dd/current/PDS4_PDS_DD_1G00.html#N-223121635

Thanks, Mark

qchaupds:
@msbentley I am still foggy.  Given this example, tell me which field identify the name of the table or array:

{pds-dev3.jpl.nasa.gov}/home/qchau/sandbox/validate 109 % view src/test/resources/github343/test_data/science_index_good.xml

     80     <Table_Character>
[...]

Issue #217, opened 2020-04-17
As a user, I want to validate content for all possible PDS4 table types


jordanpadams:
**Is your feature request related to a problem? Please describe.**
Right now, Validate only handles a subset of table types, but we need to make sure we check all possible objects.

This is a follow-on to #149 

See **referenced from** section under each of these for the classes we need examples and test cases for.

- [x] Table Character - https://pds.nasa.gov/datastandards/documents/im/v1/index_1D00.html#9.37%C2%A0%C2%A0class_pds_table_character
    - File_Area_Ancillary
    - File_Area_Browse
    - File_Area_Metadata
    - File_Area_Observational
    - File_Area_Observational_Supplemental
    - File_Area_Update

- [x] Transfer Manifests - https://pds.nasa.gov/datastandards/documents/im/v1/index_1D00.html#class_pds_transfer_manifest
- File_Area_Transfer_Manifest

- [x] Table Binary - https://pds.nasa.gov/datastandards/documents/im/v1/index_1D00.html#class_pds_table_binary
    - File_Area_Ancillary
    - File_Area_Browse
    - File_Area_Observational
    - File_Area_Observational_Supplemental

- [x] Table Delimited - https://pds.nasa.gov/datastandards/documents/im/v1/index_1D00.html#class_pds_table_delimited
    - File_Area_Ancillary
    - File_Area_Browse
    - File_Area_Metadata
    - File_Area_Observational
    - File_Area_Observational_Supplemental
    - File_Area_Update

rchenatjpl:
Here are test files. In each set of 3, the ones with records=2 or 4 should fail while the one with records=2 should pass. Filenames: XXXYYYN.xml, where
XXX =
  asc: Table_Character
  tra: Transfer_Manifest
  bin: Table_Binary
  del: Table_Delimited
YYY =
  anc: File_Area_Ancillary
  bro: File_Area_Browse
  met: File_Area_Metadata
  obs: File_Area_Observational
  sup: File_Area_Observational_Supplemental
  upd: File_Area_Update 
  inv: File_Area_Inventory
N = value of <records>. All data files have 3 records

These cases are covered, based on Jordan's IM list:
  Table Character: anc bro met obs sup
  Table Binary: anc bro obs sup
  Table Delimited: anc bro met obs sub inv

Not covered: XXXsupN because File_Area_Update is deprecated
Not covered: traYYYN. Does someone have a sample? I don't work with NSSDC
[...]

Issue #190, opened 2020-02-25
Validation fails to catch real value in ASCII_NonNegative_Integer field (Table_Delimited)


msbentley:
**Describe the bug**
Validate successfully passes the attached file, in which a real value is used in an ASCII_NonNegative_Integer field in a delimited table. Attempting to open the product e.g. in the python PDS4 reader fails because of this.

** Version of Software Used**
1.20.0

**Additional context**

[validation_test.zip](https://github.com/NASA-PDS-Incubator/validate/files/4249428/validation_test.zip)

PR #640, opened 2023-05-10
Updates for including data object names in error messages and enhance test suite with additional table types

al-niessner:
## 🗒️ Summary
Brief summary of changes if not sufficiently described by commit messages.

## ⚙️ Test Data and/or Report
One of the following should be included here:
* Reference to regression test included in code (preferred wherever reasonable)
* Attach test data here + outputs of tests

## ♻️ Related Issues
#534
#343 
#217 
#190


src/test/resources/features/validate.feature
```diff
@@ -205,6 +207,8 @@ Scenario Outline: Execute validate command for tests below.
 # https://github.com/NASA-PDS/validate/issues/188 As a user, I want to validate a bundle that uses multiple versions of the Information Model / Discipline LDDs
  |"NASA-PDS/validate#188 VALID" | "github188" | 0 | "0 errors message expected" | "totalErrors" | "src/test/resources" | "target/test" | "--skip-content-validation -r {reportDir}/report_github188_label_valid_both.json  -s json -t {resourceDir}/github188/bundle_cassini-huygens-coradar.xml {resourceDir}/github188/BILQH07S314_D065_T008S02_V02_without_Missing_Area_tag.xml" | "report_github188_label_valid_both.json" |
 
+ |"NASA-PDS/validate#217 Success with Errors: Tables Types" | "github217" | 15 | "15 error message expected" | "totalErrors" | "src/test/resources" | "target/test" | "-r {reportDir}/report_github217.json  -s json -t {resourceDir}/github217" | "report_github217.json" |
```
jordanpadams, line 210: @al-niessner can we break out the success use cases vs. the failures just so we are more explicit?

jordanpadams (changes requested):
See requested change to validate.feature

src/test/resources/features/validate.feature
al-niessner: @jordanpadams 

Can but it is going to make the features file even more ungainly. Cannot wildcard in cucumber so go from one line to 20+ lines. This has been fixed for a while so not sure this is insufficient. They all share the same tables so even breaking up the directory to keep using validates directory processing would be odd at best for the next person who needs to look at it.

Really want it broken apart?

jordanpadams: @al-niessner copy that. let's keep it as-is for now and we can break it out on a case-by-case basis down the road if specific use cases prove troublesome for whatever reason

jordanpadams (approved):

Issue #611, opened 2023-03-13
Missing validation of valid_maximum and valid_minimum from Special_Constants


rchenatjpl:
### Checked for duplicates

Yes - I've already checked

### 🧑‍🔬 User Persona(s)

data engineers

### 💪 Motivation

Does validate look at Special_Constants/valid_maximum and _minimum? It doesn't seem to, and I can imagine that would be difficult given those fields' type.

Validate passes the attached even though some values are above valid_maximum in both fields

[Archive.zip](https://github.com/NASA-PDS/validate/files/10960786/Archive.zip)

### 📖 Additional Details

Could this possibly be related to https://github.com/NASA-PDS/validate/issues/529 ?

### Acceptance Criteria

**Given** 
**When I perform** 
**Then I expect**

al-niessner:
In #529 the max/min values are defined via:

```
      <Object_Statistics>
        <maximum>0.113922</maximum>
        <minimum>0.007821</minimum>
        <mean>0.064969</mean>
        <standard_deviation>0.011258</standard_deviation>
      </Object_Statistics>
```

while in this ticket is being defined:

```
          <Special_Constants>
            <valid_maximum>444444444</valid_maximum>
          </Special_Constants>
```

While they both set a max/min value they are done differently. Tracing it back now but probably not overly related.

--- more

As suspected, when validate tries to do the min/max check for a field, they are showing up as null. This may be a parser problem and not a validate problem. Continuing to track the root cause.

al-niessner:
@jordanpadams @nutjob4life @tloubrieu-jpl 

The problem starts here:

https://github.com/NASA-PDS/validate/blob/59a6edc2aac9f739aec699c363ae3545d34b866a/src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java#L311

The start of it that the value is passed as a string. Each of the special constants are checked for equality including max/min here:

https://github.com/NASA-PDS/validate/blob/59a6edc2aac9f739aec699c363ae3545d34b866a/src/main/java/gov/nasa/pds/tools/validate/content/array/ArrayContentValidator.java#L362-L371

The impulse is to use String.compareTo() to handle the greater-than/less-than for max/min respectively. Numbers as strings do not work well in this situation because "12" is less than "2" while 2 is obviously less than 12.

[...]

PR #678, opened 2023-08-05
Update special constant min/max excursion error to be warning

### LINKED ISSUE #673: Validate does not handle Special_Constants valid_minimum and valid_maximum in accordance with information model
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

Running validate on labels with valid_minimum and valid_maximum specified results in validation errors describing values are beyond the valid_minimum and valid_maximum. However, the information model states that these values are not errors, but instead have a "special meaning". That would indicate that these validation errors are incorrect.

### 🕵️ Expected behavior

I would expect values that are beyond the valid_minimum and valid maximum to not give an error, since they are allowed to appear in the data. There might be an argument for raising a warning if the data values are beyond are beyond the valid range, however, even this is problematic, since the current validation is checking the stored value. The definitions for valid_minumum and valid_maximum do not specify whether they apply to the stored or scaled value, so this would require the attention of DDWG to fix.

I previously raised an issue about minimum and maximum values which was related to Object_Statistics, and not special constants. This seems like the appropriate place to have this validation. It would be a good idea to make the distinction between the min/max values in Object_Statistics and Special_Constants clearer, perhaps in the DPH.

https://github.com/NASA-PDS/validate/issues/434

### 📜 To Reproduce

Run validate on a label with valid_minimum and valid_maximum supplied, and a data file that contains values beyond these ranges (again, this is up for interpretation)

Sample data attached:

[grf99034.xml.zip](https://github.com/NASA-PDS/validate/files/12147994/grf99034.xml.zip)
[grf99034.fit.zip](https://github.com/NASA-PDS/validate/files/12147996/grf99034.fit.zip)

Validator output:

[grf99034.txt](https://github.com/NASA-PDS/validate/files/12148000/grf99034.txt)

### 🖥 Environment Info

- Version of this software validate 3.2.0
- Operating System: MacOS 13.4.1
...

```
gov.nasa.pds:validate
Version 3.2.0
Release Date: 2023-04-14 00:53:23

All rights reserved.
```

The definition of valid_maximum

https://pds.nasa.gov/datastandards/documents/im/v1/index_1K00.html#attribute_pds_special_constants_pds_valid_maximum

This was mentioned in the discussion for the original issue #611, however, is was after the implemenation of this feature, and should be elevated to a full issue.

https://github.com/NASA-PDS/validate/issues/611#issuecomment-1581750925
https://github.com/NASA-PDS/validate/issues/611#issuecomment-1583081886

### 🦄 Related requirements

🦄 #611

### ⚙️ Engineering Details

_No response_

al-niessner:
## 🗒️ Summary
Added new problem types to rename the out range errors to make them warning since the errors are reported correctly with same problem type. Change error messages to warnings. Cleaned up test system

## ⚙️ Test Data and/or Report
See automated tests below

## ♻️ Related Issues
Closes #673

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/ProblemType.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/ProblemType.java
@@ -117,6 +117,7 @@ public enum ProblemType {
   ARRAY_VALUE_OUT_OF_DATA_TYPE_RANGE("error.array.value_out_of_data_type_range"),
 
   ARRAY_VALUE_OUT_OF_MIN_MAX_RANGE("error.array.value_out_of_min_max_range"),
+  ARRAY_VALUE_OUT_OF_SPECIAL_CONSTANT_MIN_MAX_RANGE("warning.array.value_out_of_special_constant_min_max_range"),
 
   ARRAY_VALUE_OUT_OF_SCALED_MIN_MAX_RANGE("error.array.value_out_of_scaled_min_max_range"),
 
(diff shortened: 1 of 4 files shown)


al-niessner:
@jordanpadams 

Before #611 there was no message. Would be more appropriate to make these info rather than warning even?

jordanpadams:
@al-niessner lets throw warnings for now. I still think we should let them know in case they care.

@jstone-psi thoughts?

jordanpadams (approved):

PR #687, opened 2023-08-22
Bug fix for checking delimited tables that define an object_length

### LINKED ISSUE #684: Validation failures are contingent on presence of `<file_size>` attribute in `<File>` class
### Checked for duplicates

Seems closely related to #616, but the problem persists, despite closure of that issue.

### 🐛 Describe the bug

@katecrombie and I have been unable to figure this out. I am not sure if it's a bug with validate, or the IM, or what, but we figured this was a good place to start.

The calibration software for one of the Lucy mission's instruments uses plain-text calibration files that contain a variety of parameters and coefficients used by the processing pipeline. The file contents are organized in a reliable/well-structured way, and can be fully described by a series of `Table_Delimited` and `Stream_Text` classes in a PDS4 label.

However, if an overall `<file_size>` attribute is is included in the `<File>` class (as is standard practice), the label fails validation due to the following error(s):
```
ERROR  [error.label.invalid_object_definition]   line 2: Invalid object definition for object #2.  The previously defined object ends at byte 94. Offset defined in label: 24
....
```

Each individual data object (`Table_Delimited` or `Stream_Text`) has a correctly-defined `<offset>` and `<object_length>`, and simply omitting the `<file_size>` enables it to pass validation without complaint.

### 🕵️ Expected behavior

I expected that inclusion/omission of the `<file_size>` attribute in the `<File>` class in the `<File_Area_Observational>` section would not have an impact on validation failure/passage.

### 📜 To Reproduce

1. Create an XML label for a product that contains several `Table_Delimited` and/or `Stream_Text` objects
2. In the `<File>` class, include a `<file_size>` attribute (with `unit="byte"`) with a value equal to the product's file size, in bytes
3. Run `validate`, and note failure
4. Remove or comment out the `<file_size>` attribute from the label completely
5. Re-run `validate`, and note passing result

### 🖥 Environment Info

Operating System: Issue replicated on both MacOSX and Ubuntu 20.04

`validate` versions 3.1.1 and 3.2.0

[Attached is a zip file](https://github.com/NASA-PDS/validate/files/12381311/filesize_validation_bug.zip) containing a dummy data product of the same format as described, with two corresponding labels whose only differences are the minor version number of their VID (so that both can be validated simultaneously), and whether or not the `<file_size>` attribute is present on line 61. A validation report showing that one passes while the other fails is also included.

al-niessner:
## 🗒️ Summary
Unit tests are here but the actual fix is in NASA-PDS/pds4-jparser#97

## ⚙️ Test Data and/or Report
See automated unit tests

## ♻️ Related Issues
Closes #684


al-niessner:
@jordanpadams 

jparser PR be merged and maven artifact updated for these unit tests to pass assuming I got all the cucumber stuff right.

jordanpadams:
Merged PR on pds4-jparser. Rerunning branch tests.

nutjob4life (approved):
lgtm

PR #688, opened 2023-08-22
More concise error message when possible record_length mismatch

### LINKED ISSUE #680: validate does not correctly diagnose errors in the record_length value for fixed width Table_Character objects
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

For Table_Character labels containing a single error of an incorrect value for:

File_Area_Observational/Table_Character/Record_Character/record_length

Validate reports a large number of different errors, including multiple instances of records not ending in the indicated record delimiter and data values not matching the data type indicated in the label. 

### 🕵️ Expected behavior

While the reported errors are technically true if Validate is reading the number of bytes reported in the record_length, they are not reporting the real problem. It would be better if Validate reported the actual error of the incorrect record_length. 

### 📜 To Reproduce

1. Take a valid PDS4 label containing a Table_Character object, and change the Record_Length value.
2. Run Validate tool on the modified label and compare the validation report to that of the original label.

...
I've attached a zip file containing a sample data file with two PDS4 labels. ORB12_EUR_EPHIO_reclen96.xml is the original and has the correct record_length value of "96". The other, ORB12_EUR_EPHIO_reclen95.xml, has the record_length value changed to "95". The error report produced by Validate Tool is also included.

[ORB12_EUR_EPHIO.zip](https://github.com/NASA-PDS/validate/files/12363309/ORB12_EUR_EPHIO.zip)

### 🖥 Environment Info

- Validate version: 3.2.0
- Operating System: Linux (Ubuntu 20)
...

V.3.2.0 (2023-04-14 00:53:23)

_No response_

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

al-niessner:
## 🗒️ Summary
In the condition of tables with fixed record lengths and not enough data to fill all of the fields and record delimiter, then have one error and move onto next data object rather than the current 50000+ errors that now show for even the smallest of tables.

## ⚙️ Test Data and/or Report
See automated unit tests below

## ♻️ Related Issues
Closes #680


nutjob4life (approved):
👍

PR #691, opened 2023-08-30
Update `-u` functionality for updates context products to use new PDS Search API

### LINKED ISSUE #675: Cutover to using new Registry API for generating context products json
## 💡 Description

Update `-u` flag functionality to query API instead of legacy registry.

al-niessner:
## 🗒️ Summary
Instead of using Solr to read the old database, use simple http to read from an registry-api instead.

## ⚙️ Test Data and/or Report
NA - run validate -u and look at the JSON it generates

## ♻️ Related Issues
Closes #675

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
+++ b/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
@@ -40,7 +40,9 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLEncoder;
 import java.nio.channels.FileChannel;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


al-niessner:
Notes:

1. because of pagination, takes forever to run just up to the 10000
2. because registry-api does not have 10000 fix so when that is applied it will take longer
3. odd difference between curl and Java that do not get expected registry file (empty with java while filled with curl)
    a. same URL (double checked)
    b. when fed curl version of data, Java produces expected file
    c. hence, it really is something weird between Java and curl

nutjob4life (approved):

jordanpadams:
@al-niessner based upon your comment above, does that mean we actually do not get all the context products? Or are there more than 10000?

jordanpadams:
@al-niessner ☝️

jordanpadams:
Pausing the merger of this PR until:
1. we can guarantee the accuracy of the context products loaded.
2. Pagination performance improves

Issue #335, opened 2021-05-04
validate gives a NullPointerException during validation of a directory containing Table_Character products


msbentley:
## 🐛 Describe the bug
I am getting the following output when trying to validate a directory of products using `-t .` in a directory containing Table_Character products. This always used to work, but now I get errors like:
```
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at gov.nasa.pds.tools.validate.rule.AbstractValidationRule.execute(AbstractValidationRule.java:71)
	at org.apache.commons.chain.impl.ChainBase.execute(ChainBase.java:191)
	at gov.nasa.pds.tools.validate.rule.pds4.LabelInFolderRule$1.run(LabelInFolderRule.java:154)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.NullPointerException
	at gov.nasa.pds.tools.validate.rule.RuleContext.getCheckInbetweenFields(RuleContext.java:427)
	at gov.nasa.pds.tools.validate.rule.pds4.TableDataContentValidationRule.getCheckInbetweenFields(TableDataContentValidationRule.java:115)
	at gov.nasa.pds.tools.validate.rule.pds4.TableDataContentValidationRule.validateTableDataContents(TableDataContentValidationRule.java:397)
	... 12 more
```
note that if I explicitly use `-t *.xml` then it works OK!

The following test product (uploaded also for another ticket) shows the problem:

[validate_field_formats.zip](https://github.com/NASA-PDS/validate/files/6420240/validate_field_formats.zip)

[...]

PR #721, opened 2023-10-04
Fix bug where precision is not calculated correctly with whitespace in field

### LINKED ISSUE #681: Validate incorrectly throws precision mismatch error for Table_Delimited
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When performing content validation on a delimited table, Validate throws a field_value_format_precision_mismatch error if any value in a field has less precision than dictated by field_format.

### 🕵️ Expected behavior

Per StdRef 4B.1.2, field_format defines the maximum precision, not the only precision allowed:

> For character tables, <field_format> is used to describe the maximum length and alignment of
> the data. <field_format> also gives an indication of the maximum precision of real numbers, but
> does not require all values to have this precision.

Validate does correctly not throw an error if the same table is described by a Table_Character object.

StdRef 4C.2, which describes delimited tables, states:

> Values for attribute <field_format> are set as described in Section 4B.1.2.

...so the same behavior should apply.

### 📜 To Reproduce

Run Validate on the included labels. Check the validation report.

### 🖥 Environment Info

- Operating System: Windows 11 Pro

Validate Version 3.2.0

Included in the .zip file are:

Table_Delimited label
Table_Character label
CSV table that produces the error for delimited but not character
CSV table with whitespace removed that exhibits the same behavior (just in case)
Validation reports showing the error

[ff_test.zip](https://github.com/NASA-PDS/validate/files/12363982/ff_test.zip)

### 🦄 Related requirements

_No response_

### ⚙️ Engineering Details

_No response_

### I&T
TestRail Test ID: T8681195

al-niessner:
## 🗒️ Summary
Added check that if delimited table has trailing spaces on number to flag it as data type mismatch error instead of precision.

## ⚙️ Test Data and/or Report
See automated unit tests below (should pass)

## ♻️ Related Issues
Closes #681


jordanpadams (changes requested):
@al-niessner one thing we can't forget here is in a delimited table, everything within the delimiters is considered part of the "field value", making spaces within any numeric field invalid.

This only applies to delimited tables. You can pad in character tables since they are fixed width fields.

From the Standards Reference:
Screenshot 2023-10-04 at 10 36 55 AM

jordanpadams:
For example if I prepend a space in the field for this CSV, it passes, but it should fail because `\s2.22` is not an `ASCII_Real`

[ff_test.csv](https://github.com/NASA-PDS/validate/files/12806537/ff_test.csv)

nutjob4life (approved):
LGTM 👍

al-niessner:
@jordanpadams done

src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
```diff
@@ -840,13 +840,20 @@ private void checkFormat(String value, String format, int fieldIndex,
...
           }
           if (length != precision) {
-            addTableProblem(ExceptionType.ERROR, ProblemType.FIELD_VALUE_FORMAT_PRECISION_MISMATCH,
-                "The number of digits to the right of the decimal point " + "in the value '"
-                    + value.trim() + "' does not equal the "
-                    + "precision set in the defined field format " + "(expected " + precision
-                    + ", got " + length + ").",
-                recordLocation, fieldIndex);
             isValid = false;
+            if (value.trim().length() == value.length()) {
```
jordanpadams: @al-niessner This fixes the whitespace padding when a `field_format` or `validation_format` are specified, but it looks like this may overall be a more inherent bug in handling of whitespace in delimited tables. When I remove the `field_format` attributes from the `ff_del.xml` label, it passes, but it should fail. I created a new requirement here that ties to this handling of whitespace in tables more explicitly: https://github.com/NASA-PDS/validate/issues/722

I would just merge and then work on that requirement, but I think this solution will immediately be changed based upon that requirement

jordanpadams (changes requested):
See review comment

src/main/java/gov/nasa/pds/tools/validate/content/table/FieldValueValidator.java
al-niessner: done

al-niessner:
@jordanpadams 

Oh my. This really breaks a bunch of old tests and specifically #335 that is testing character/delimited tables that are the same. The table in 335 is:

```
2019-08-06T00:00:00Z, This is a test, 1111, 2222, 3333, 4444
2019-08-06T00:01:00Z, This is a test, 1111, 2222, 3333, 4444
2019-08-06T00:02:00Z, This is a test, 1111, 2222, 3333, 4444
2019-08-06T00:03:00Z, This is a test, 1111, 2222, 3333, 4444
````

From XML delimiter is Comma. I do not know if there is a Comma-Space but without it, this file breaks. Does this mean we have been broken all along or that we are going the wrong direction?

jordanpadams:
@al-niessner hold off on any further implementation on this ticket. we have some comments from people requiring discussion on the standards.

jordanpadams:
@al-niessner 
> Does this mean we have been broken all along or that we are going the wrong direction?

my brain very well may be taking us in the wrong direction. less pause on this for now. thanks for checking this out.

nutjob4life (approved):

jordanpadams (approved):

jordanpadams (approved):

jordanpadams (approved):
🎉

jordanpadams:
CI tests are failing due to intermittent network blocking/outage. But tests run successfully on occasion.

PR #726, opened 2023-10-06
Upgrade net.sf.saxon:Saxon-HE from 9.9.1-7 to 12.3

### LINKED ISSUE #707: Upgrade to Saxon 12.x
## 💡 Description

[Saxon 9 appears to be EOL](https://github.com/Saxonica/Saxon-HE#current-releases), so let's upgrade Saxon to the latest version. Per this [branch test from dependabot PR](https://github.com/NASA-PDS/validate/actions/runs/6355118777), there are some backwards incompatible updates we will need to address.

Changes can be made directly to this dependabot branch: https://github.com/NASA-PDS/validate/tree/dependabot/maven/net.sf.saxon-Saxon-HE-12.3

- [x] Update for initial merge here: https://github.com/NASA-PDS/validate/pull/701
- [ ] Update next to catch erroneous messages hitting stdout: https://github.com/NASA-PDS/validate/pull/701#issuecomment-1749265874

al-niessner:
## 🗒️ Summary
The new SAXON libraries are more pedantic about processing XML files. It now generates warning and errors that cannot be squashed - well I could not find the wire to cut to the blinking red light. Instead, changed validate so that it would not ask SAXON to parse files that do not have the XML extension given by the user or defined by default.

## ⚙️ Test Data and/or Report
All automated unit tests should pass

## ♻️ Related Issues
Closes #707

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
+++ b/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
@@ -22,6 +22,7 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.util.HashMap;
 import java.util.Map;
+import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.transform.Source;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerConfigurationException;
@@ -60,13 +61,14 @@ public class SchematronTransformer {
     // Use saxon for schematron (i.e. the XSLT generation).
     System.setProperty("javax.xml.transform.TransformerFactory",
         "net.sf.saxon.TransformerFactoryImpl");
+    DocumentBuilderFactory.newInstance().setNamespaceAware(true);
     TransformerFactory isoFactory = TransformerFactory.newInstance();
     // Set the resolver that will look in the jar for imports
     isoFactory.setURIResolver(new XslURIResolver());
     // Load the isoSchematron stylesheet that will be used to transform each
     // schematron file
     Source isoSchematron = new StreamSource(
-        LabelValidator.class.getResourceAsStream("/schematron/iso_svrl_for_xslt2.xsl"));
+        LabelValidator.class.getResourceAsStream("/schematron/simple_test.xsl"));
     isoTransformer = isoFactory.newTransformer(isoSchematron);
     transformerFactory = TransformerFactory.newInstance();
     cachedTransformers = new HashMap<String, Transformer>();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [«unknown»](https://github.com/NASA-PDS/validate/tree/«unknown») (2023-10-05)
+## [«unknown»](https://github.com/NASA-PDS/validate/tree/«unknown») (2023-10-17)
 
 [Full Changelog](https://github.com/NASA-PDS/validate/compare/v3.3.0...«unknown»)
 
@@ -10,6 +10,7 @@
 
 **Defects:**
 
+- Installation instructions need updates for Windows [\#723](https://github.com/NASA-PDS/validate/issues/723)
 - Investigate and Fix code scanning alert for potential improvements [\#698](https://github.com/NASA-PDS/validate/issues/698) [[s.medium](https://github.com/NASA-PDS/validate/labels/s.medium)]
 - Validate crashes on invalid \# of records, and spurious file left behind [\#475](https://github.com/NASA-PDS/validate/issues/475) [[s.low](https://github.com/NASA-PDS/validate/labels/s.low)]
 
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -179,7 +179,7 @@ recommend that a file or class name and description of purpose be included on
 the same “printed page” as the copyright notice for easier identification within
 third-party archives.
 
-    Copyright [yyyy] [name of copyright owner]
+    Copyright 2023 California Institute of Technology
     
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,6 +1,6 @@
 PDS Validate Tool
 
-Copyright 2020–2021, California Institute of Technology ("Caltech").
+Copyright 2020–2023, California Institute of Technology ("Caltech").
 U.S. Government sponsorship acknowledged.
 
 All rights reserved.
--- a/pom.xml
+++ b/pom.xml
@@ -102,7 +102,7 @@ POSSIBILITY OF SUCH DAMAGE.
     <plugins>
        <plugin>
           <artifactId>maven-surefire-plugin</artifactId>
-          <version>3.0.0-M3</version>
+          <version>3.1.2</version>
           
           <configuration>
           <argLine>-Dorg.slf4j.simpleLogger.defaultLogLevel=ERROR</argLine>
@@ -200,7 +200,7 @@ POSSIBILITY OF SUCH DAMAGE.
     <dependency>
       <groupId>net.sf.saxon</groupId>
       <artifactId>Saxon-HE</artifactId>
-      <version>12.3</version>
+      <version>9.9.1-7</version>
     </dependency>
     <dependency>
       <groupId>commons-cli</groupId>
@@ -315,7 +315,7 @@ POSSIBILITY OF SUCH DAMAGE.
     <dependency>
       <groupId>org.apache.solr</groupId>
       <artifactId>solr-solrj</artifactId>
-      <version>7.7.2</version>
+      <version>9.3.0</version>
     </dependency>
     <dependency>
         <groupId>org.junit.jupiter</groupId>
--- a/src/main/java/gov/nasa/pds/tools/label/LabelValidator.java
+++ b/src/main/java/gov/nasa/pds/tools/label/LabelValidator.java
@@ -84,7 +84,7 @@ import gov.nasa.pds.tools.validate.ProblemType;
 import gov.nasa.pds.tools.validate.TargetExaminer;
 import gov.nasa.pds.tools.validate.ValidationProblem;
 import gov.nasa.pds.validate.constants.Constants;
-import net.sf.saxon.om.TreeInfo;
+import net.sf.saxon.om.DocumentInfo;
 import net.sf.saxon.trans.XPathException;
 
 /**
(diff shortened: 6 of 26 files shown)


al-niessner:
@jordanpadams 

These warning still persists:
`WARNING: Supplied DOM uses namespaces, but is not created as namespace-aware`

Cannot make it go away but will keep poking at it until you ask for something else to be done.

al-niessner:
@jordanpadams 

Very interesting, Schematron error is back but on different XML files. This time it did on an non-aggregate label. Super strange because it is implying that it the PDS host that does not always server the schematron. I ran 651 locally and it did what was expected. The other 100+ also load the schematron while 651 is near the top of the list and 271 is near the bottom. What would you like to do?

jordanpadams:
@al-niessner I will take a look. My guess is our JPL or AWS are blocking GitHub because we are pinging it so often.

jordanpadams:
@al-niessner per this:
> WARNING: Supplied DOM uses namespaces, but is not created as namespace-aware
We definitely need to figure out how to suppress or capture/dump those warnings if at all possible? Or maybe even hack it by have the `java` execution in the bash script send `2>/dev/null`?

al-niessner:
@jordanpadams 

Finally found the cause of the `WARNING: Supplied DOM uses namespaces, ...` and it is way down in Apache code that uses the SAXON library that we now use. It happens when we read validation-commands.xml. Maybe some others but get plural same warning on reading that one file. I am trying to figure out if there is something I can add to the XML that helps solve it like a default namespace or something.

Cannot run it in a different thread and send results to /dev/null because we need the results of the parse in memory.

The reason for the warning is XPath. SAXON moved to XPath 2 which now requires valid namespace usage so having an XML with namespace but a DOM without it and then use XPath, it gives this warning because it cannot be done.

jordanpadams:
@al-niessner copy that. We may be able to throw together a quick a dirty schema for this if it really needs one.

jordanpadams:
@al-niessner actually just did using https://www.freeformatter.com/xsd-generator.html:
```
<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:element name="catalog">
    <xs:complexType>
      <xs:choice maxOccurs="unbounded" minOccurs="0">
        <xs:element name="chain">
          <xs:complexType>
            <xs:sequence>
              <xs:element name="command" maxOccurs="unbounded" minOccurs="0">
                <xs:complexType>
                  <xs:simpleContent>
                    <xs:extension base="xs:string">
                      <xs:attribute type="xs:string" name="caption" use="optional"/>
                      <xs:attribute type="xs:string" name="className" use="optional"/>
                    </xs:extension>
                  </xs:simpleContent>
                </xs:complexType>
              </xs:element>
            </xs:sequence>
            <xs:attribute type="xs:string" name="name"/>
            <xs:attribute type="xs:string" name="caption"/>
            <xs:attribute type="xs:string" name="className"/>
          </xs:complexType>
        </xs:element>
        <xs:element name="command">
          <xs:complexType>
            <xs:sequence>
              <xs:element name="command" maxOccurs="unbounded" minOccurs="0">
                <xs:complexType>
                  <xs:simpleContent>
                    <xs:extension base="xs:string">
                      <xs:attribute type="xs:string" name="caption" use="optional"/>
                      <xs:attribute type="xs:string" name="className" use="optional"/>
                    </xs:extension>
                  </xs:simpleContent>
                </xs:complexType>
              </xs:element>
            </xs:sequence>
            <xs:attribute type="xs:string" name="name" use="optional"/>
            <xs:attribute type="xs:string" name="caption" use="optional"/>
            <xs:attribute type="xs:string" name="className" use="optional"/>
          </xs:complexType>
        </xs:element>
      </xs:choice>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

Maybe we can just drop that schema alongside the file and bind it to XML as it is read?

al-niessner:
@jordanpadams 

Finally, made it to the end. Still cannot fix it. Here is what is going on:

It all starts here as this called many times always and only causing the warning:
https://github.com/NASA-PDS/validate/blob/eb8f0f962be8e002a420770afc74290fb60eddfb/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java#L70

It looks as this file is being read in when it does the import:
https://github.com/NASA-PDS/validate/blob/eb8f0f962be8e002a420770afc74290fb60eddfb/src/main/resources/schematron/iso_svrl_for_xslt2.xsl#L155

It is during the parsing of this file, presumably the previous file has been on hold and will have the same problem when it hits this bit of SAXON code:
```
                    if (top.getNodeName().contains(":")) {
                        issueWarning(receiver, "Supplied DOM uses namespaces, but is not created as namespace-aware");
                    } else if (top.hasAttribute("xmlns")) {
                        issueWarning(receiver, "Supplied DOM declares a default namespace, but is not created as namespace-aware");
                    }
```

where top is `[xsl:stylesheet: null]`. Some clue, but not a good enough one, is given in the comments about testing these conditions:
```
            // Do a quick check (not guaranteed to catch all cases) that this isn't a non-namespace-aware DOM,
            // to catch the common case of failing to set namespace-awareness on the DOM factory
```

I cannot find where to set the namespace awareness from the code in SchematronTransfomer. They did a good masking it from simple searches and google did not give me enough with 'namespace aware TranformerFactory'.

The interesting part is that it calls `issueWarning` which looks like this:
```
    private static void issueWarning(Receiver receiver, String text) {
        try {
            Configuration config = receiver.getPipelineConfiguration().getConfiguration();
            config.getLogger().warning("WARNING: " + text);
        } catch (Exception err) {
            System.err.println("WARNING: " + text);
        }
    }
```
If we (I) could figure out how to configure their logger, it would not show up on stderr. However, that is immaterial as it should be possible to fix the transformer to be namespace aware.

I went to do the 2> /dev/null in validate script but if we do that it will hide all errors if java fails for any reason making it hard on those are just starting with validate. Those that have a good environment now will not notice, In fact, those with a working environment will not upgrade. Still I made the changes but we should put aside some time to revisit this hack and fix it properly in SchematronTransformer.

jordanpadams:
@al-niessner can we try to set the pipeline configuration or know what that exception is? Could we maybe [clone/fork that repo](https://saxonica.plan.io/projects/saxon/repository) and modify the code to see what happens?

We can also file a bug and maybe even a PR on their website to address this. We have submitted PRs to them in the past.

jordanpadams:
If nothing else, by building the code locally, you can print the stack trace to see why it is failing

al-niessner:
> @al-niessner can we try to set the pipeline configuration or know what that exception is? Could we maybe [clone/fork that repo](https://saxonica.plan.io/projects/saxon/repository) and modify the code to see what happens?
> 
> We can also file a bug and maybe even a PR on their website to address this. We have submitted PRs to them in the past.

@jordanpadams 

It is not an exception. When SAXON finds an `<xsl:import>` it gives us this warning. For reasons that are unknown to me, during the import it does a test to see if the imported xsl has namespacing (':' in the node name), which it does, but the DOM object was not created with namespaces. SAXON does not read the xmlns from the declaration.

The exception handling is just a cheesy way of getting the message to stderr if logging has not been initiated.

jordanpadams:
@al-niessner Copy that. Can we file a bug in their system? We then may need to table this PR for the time being until we can get this resolved. the more I think about the `2> /dev/null` workaround, the more I think that is a bad idea... We will lose all traceability from a user when something goes wrong.

al-niessner:
> @al-niessner Copy that. Can we file a bug in their system? We then may need to table this PR for the time being until we can get this resolved. the more I think about the `2> /dev/null` workaround, the more I think that is a bad idea... We will lose all traceability from a user when something goes wrong.

@jordanpadams 

Agreed on 2>.

I was wondering if it was our problem - as in something with the xsl files - so went to W3 Schools and got their example for xsl:import. Gives the same warning. Been toying with saxon developer site trying to figure out how to post something to their forum as they do not provide a subscribe button. Once I do, I will post the W3 Schools example and ask if it is their bug, our bug, or live with bug (aka feature).

jordanpadams:
@al-niessner 👍 

Other super hacky way we could do this is something like:
```
$ validate ... 2> .tmp.log
$ grep -v "WARNING: Supplied DOM uses namespaces" .tmp.log
$ rm -f .tmp.log
```
But not idea how we would do that on windows...

al-niessner:
> @al-niessner 👍
> 
> Other super hacky way we could do this is something like:
> 
> ```
> $ validate ... 2> .tmp.log
> $ grep -v "WARNING: Supplied DOM uses namespaces" .tmp.log
> $ rm -f .tmp.log
> ```
> 
> But not idea how we would do that on windows...

@jordanpadams 

The super hacky way I tried but tired of making it work before reaching the goal line was to manually import skeleton with svrl, removing the xsl:import from svrl, into all_in_one.xsl. I figured that could let us limp along with no errors until we wanted to update something in them. Turns out that acting like the import statement is more complicated than just merging the two files.

jordanpadams:
@al-niessner interesting. I think since this version doesn't add anything in particular to our software, I am going to revert the previous PR and keep this PR in a holding pattern until we can get a fix from them (or a security issue pushes us to upgrade anyways).

nutjob4life:
> But not idea how we would do that on windows...

I think it's:

    Get-Content .\.tmp.log | Select-String -Pattern "WARNING: Supplied DOM uses namespaces" -NotMatch

al-niessner:
Follow [saxon help](https://saxonica.plan.io/boards/3/topics/9551)

al-niessner:
@jordanpadams 

It fails because it has merged main back into this branch. I can force push my version of this branch which will revert all of the changes you undid on main to this branch and it should work. In theory, it would be a one step merge of this branch to main to move to saxon 12. Let me know what you want to do.

jordanpadams:
@al-niessner you are a go for force push

src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
```diff
@@ -60,13 +61,14 @@ public SchematronTransformer() throws TransformerConfigurationException {
...
         "net.sf.saxon.TransformerFactoryImpl");
+    DocumentBuilderFactory.newInstance().setNamespaceAware(true);
     TransformerFactory isoFactory = TransformerFactory.newInstance();
     // Set the resolver that will look in the jar for imports
     isoFactory.setURIResolver(new XslURIResolver());
     // Load the isoSchematron stylesheet that will be used to transform each
     // schematron file
     Source isoSchematron = new StreamSource(
-        LabelValidator.class.getResourceAsStream("/schematron/iso_svrl_for_xslt2.xsl"));
+        LabelValidator.class.getResourceAsStream("/schematron/simple_test.xsl"));
```
jordanpadams, line 71: @al-niessner just as a heads up, I think this was left behind in your commits by mistake?

jordanpadams (changes requested):

src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
al-niessner: Yes. Fixed.

al-niessner:
@jordanpadams 

Forcing does not work as this does include what was undone in main branch. Ended up creating a patch, creating a new branch, applying patch. This PR is dead and should be ignored (I am closing it) and should use PR #734

jordanpadams:
thanks @al-niessner sorry for the confusion!

PR #742, opened 2023-10-27
Fix intermittent bug with reading schematrons per Saxon upgrade

### LINKED ISSUE #739: Intermittent network failures attempting repeated downloads of schemas/schematrons
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I did a `mvn test`, it works sometimes, sometimes it fails indicating it cannot download a schema or schematron, and other times it just hangs...

Atmospheres also noted that validate is running and then just hangs and appears to not be doing anything.

### 🕵️ Expected behavior

I expected it would download and test successfully.

### 📜 To Reproduce

1. `mvn test` (may need to try it a few times)

### 🖥 Environment Info

Mac OSx (me)
Linux (ATM Node)

v3.2.0
v3.4.0-SNAPSHOT

Any large test data sets.

### 🦄 Related requirements

- 🦄 #707

### ⚙️ Engineering Details

- DSIO ticket created to investigate CloudWatch and any blocking happening there

- For development, let's verify our local caching is actually working and we are not going back to JPL website over and over again

al-niessner:
## 🗒️ Summary
Fixed a bad use of null and replaced it with a NullObject that behaves like an object returning obviously null information but preventing null pointer exceptions while handling exceptions.

## ⚙️ Test Data and/or Report
All unit tests pass

## ♻️ Related Issues
Closes #739 
Refs #707

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
+++ b/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
@@ -154,7 +154,12 @@ public class SchematronTransformer {
         in = Utility.openConnection(conn);
 
         StreamSource source = new StreamSource(in);
-        //source.setSystemId(schematron.toString());
+        /* This fixes the problem but is it a good fix. The message now indicates that it is finding
+         * the same URI (systemId) for two files. The one it is reading and the source that we have?
+         * Maybe in saxon 12.3 it fills in the systemId correctly? Cannot find anything that makes
+         * this fix definitive good or bad but that is a search skill problem not lack of information
+         * 
+         * source.setSystemId(schematron.toString()); */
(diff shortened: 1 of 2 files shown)


al-niessner:
@jordanpadams 

Maybe you understand saxon better but if you do not find an expert and ask about these changes. From the saxon error (after fixing the null pointer problem in first change) found that it thought the XLST was being added more than once with different content. All I could think of was that the source systemId is no longer needed in 12.3. Cannot find the line in the release notes to confirm. Anyway, commenting that out makes it all work for good or ill.

jordanpadams:
awesome thanks @al-niessner looks great!

jordanpadams (approved):

PR #769, opened 2023-11-21
Throw ERROR when file areas point at same file

### LINKED ISSUE #755: As a user, I want validate to throw an error when a file is being referenced by more than one label
### Checked for duplicates

Yes - I've already checked

### 🧑‍🔬 User Persona(s)

All the above; anyone that is validating collections and bundles or a directory of products.

### 💪 Motivation

According to the standards a file should only be described once by a label, and then referenced with a LID if multiple files want to point to it.  Currently validate does not check for this.  We recently found a case in the dart_teleobs:data_mroraw where a single file was described by a File_Area_Observational_Supplemental within multiple xml labels.  No one thought to check for this, but I would hope this would be easy to check for this, especially since we already check to see if each file has a label.  

Very similar in nature to #245 

### 📖 Additional Details

Example:  https://pdssbn.astro.umd.edu/holdings/pds4-dart_teleobs:data_mroraw-v1.0/mro_221011/

The file "comps_221011.png" is listed in the File_Area_Observational_Supplemental for 30 files in this directory, the first three of which listed below:

- m221011.0013.xml
- m221011.0014.xml
- m221011.0030.xml

### Acceptance Criteria

**Given** a set of label files where >1 of them references the same file from `File_Area_*`
**When I perform** `validate --target product1.xml product2.xml etc.`
**Then I expect** validate to throw an error

### ⚙️ Engineering Details

When keeping inventory of the files that have been referenced, we can drop the inventory when we move to another directory.

### I&T
TestRail Test ID: T8681202

al-niessner:
## 🗒️ Summary
Check during file area load if the file is already "in use" and report an error if it is.

## ⚙️ Test Data and/or Report
Automated unit tests below all pass

## ♻️ Related Issues
Closes #755


nutjob4life (approved):
Oh wait … it's a draft

al-niessner:
@nutjob4life 

Sorry, I think my work flow is different than the rest of ya. Since I am my own beat deaf drummer - yeah I make Elaine look like a good dancer - that should not be too much of a surprise. Anyway, I like to use the PR for testing since it works so well here which means I do a lot of pushes to it before letting it go. Sorry that it tricks you into thinking it is time for a review when it is not.

nutjob4life:
@al-niessner no worries; it's a [known and open question](https://github.com/orgs/community/discussions/10409) whether draft pull requests should send notifications.  I wish GitHub would notify only when it's non-draft.

I'll try to pay attention to the "draft" status, but once it's ready, you may have to re-ping me to get my review. (I practice "inbox zero".)

jordanpadams (approved):
I am slightly concerned with our memory utilization having another data object that is going to need to store references for every target we come across, but I guess that is a necessary evil without some complex logic to know when you can flush the data object. Let's move forward with this and see if we encounter additional issues.

PR #782, opened 2023-12-04
Add docker container deployment steps

jordanpadams:
Copied from https://github.com/NASA-PDS/registry-api/blob/main/.github/workflows/unstable-cicd.yaml

Note: Not sure if this is correct or not with the Dockerfile in this repo. Hoping for some help from reviewers.

### Testing
```
]$ docker image build --tag nasapds/validate --build-arg tar_file=target/*-bin.tar.gz --file docker/Dockerfile .
[+] Building 2.0s (8/8) FINISHED                                                                                                                       docker:desktop-linux
 => [internal] load .dockerignore                                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                                        0.0s
 => [internal] load build definition from Dockerfile                                                                                                                   0.0s
 => => transferring dockerfile: 2.10kB                                                                                                                                 0.0s
 => [internal] load metadata for docker.io/library/openjdk:15-slim                                                                                                     0.9s
 => [internal] load build context                                                                                                                                      0.1s
 => => transferring context: 165B                                                                                                                                      0.1s
 => CACHED [1/3] FROM docker.io/library/openjdk:15-slim@sha256:1e069bf1c5c23adde58b29b82281b862e473d698ce7cc4e164194a0a2a1c044a                                        0.0s
 => [2/3] COPY target/*-bin.tar.gz /tmp/validate-bin.tar.gz                                                                                                            0.1s
 => [3/3] RUN  mkdir /opt/validate       && tar xzf /tmp/validate-bin.tar.gz  -C /opt/validate --strip-components 1      && rm -f /tmp/validate-bin.tar.gz      && ap  0.7s
 => exporting to image                                                                                                                                                 0.1s
 => => exporting layers                                                                                                                                                0.1s
 => => writing image sha256:335c5562256c7af4a01e9fe27fb753c8521cb5b8ce48a0b44a3dfef95fc66118                                                                           0.0s
 => => naming to docker.io/nasapds/validate
```

### DIFF
--- a/.github/workflows/stable-cicd.yaml
+++ b/.github/workflows/stable-cicd.yaml
@@ -80,10 +80,9 @@ jobs:
                     ADMIN_GITHUB_TOKEN: ${{secrets.ADMIN_GITHUB_TOKEN}}
 
             -
-                name: 🫙 Jar and Tag Determination
-                id: jartag
-                run: |
-                    echo "jar_file=$(find ./service/target/ -maxdepth 1 -regextype posix-extended -regex '.*/validate-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?\.jar')" >> $GITHUB_OUTPUT
+                name: 🫙 Tar and Tag Determination
+                id: gettartag
+                run: echo "tar_file=$(tar_file=$(find ./target/ -maxdepth 1 -regextype posix-extended -regex '.*/*-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?\.tar.gz')" >> $GITHUB_OUTPUT
                     echo "image_tag=$(echo ${{github.ref}} | awk -F/ '{print $NF}')" >> $GITHUB_OUTPUT
             -
                 name: 💳 Docker Hub Identification
@@ -103,10 +102,10 @@ jobs:
                 with:
                     context: ./
                     file: ./docker/Dockerfile
-                    build-args: validate_jar=${{steps.jartag.outputs.jar_file}}
+                    build-args: tar_file=${{steps.gettartag.outputs.tar_file}}
                     platforms: linux/amd64,linux/arm64
                     push: true
-                    tags: ${{secrets.DOCKERHUB_USERNAME}}/validate:${{steps.jartag.outputs.image_tag}}
+                    tags: ${{secrets.DOCKERHUB_USERNAME}}/validate:${{steps.gettartag.outputs.image_tag}}
 
 ...
 
--- a/.github/workflows/stable-cicd.yaml
+++ b/.github/workflows/stable-cicd.yaml
@@ -82,7 +82,7 @@ jobs:
             -
                 name: 🫙 Tar and Tag Determination
                 id: gettartag
-                run: echo "tar_file=$(find ./target/ -maxdepth 1 -regextype posix-extended -regex '.*/*-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?\.tar.gz')" >> $GITHUB_OUTPUT
+                run: echo "tar_file=$(find ./target/ -maxdepth 1 -regextype posix-extended -regex '.*/.*-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?\.tar.gz')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


jordanpadams:
I feel like this should work, but I can't get it to test properly on my laptop.

.github/workflows/stable-cicd.yaml
```diff
@@ -79,6 +79,34 @@ jobs:
                     CODE_SIGNING_KEY: ${{secrets.CODE_SIGNING_KEY}}
                     ADMIN_GITHUB_TOKEN: ${{secrets.ADMIN_GITHUB_TOKEN}}
 
+            -
+                name: 🫙 Tar and Tag Determination
+                id: gettartag
+                run: echo "tar_file=$(find ./target/ -maxdepth 1 -regextype posix-extended -regex '.*/.*-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?-bin\.tar.gz')" >> $GITHUB_OUTPUT
```
nutjob4life: Need a `\` in front of `.gz`

.github/workflows/unstable-cicd.yaml
```diff
@@ -84,6 +84,33 @@ jobs:
                     CODE_SIGNING_KEY: ${{secrets.CODE_SIGNING_KEY}}
                     ADMIN_GITHUB_TOKEN: ${{secrets.ADMIN_GITHUB_TOKEN}}
 
+            -
+                name: 🫙 Tar Determination
+                id: gettar
+                run: echo "tar_file=$(find ./target/ -maxdepth 1 -regextype posix-extended -regex '.*/.*-[0-9]+\.[0-9]+\.[0-9]+(-SNAPSHOT)?-bin\.tar.gz')" >> $GITHUB_OUTPUT
```
nutjob4life: ditto

nutjob4life (approved):
Couple minor nits but otherwise excellent! 🎉

nutjob4life (approved):

nutjob4life (approved):

jordanpadams:
@nutjob4life any idea why the steps described in the README to build this docker image locally doesn't work?

nutjob4life:
@jordanpadams change the `ADD` to `COPY` and you're good to go!

docker/README.md
```diff
@@ -1,44 +1,30 @@
...
-worry about the Java/JDK versions installed and operating systems supported. Also, this helps to 
-easily use the Validate Tool in other execution environments such as docker compose and Amazon ECS.
+There's one Dockerfile here that can be used to make images for development and production.
 
-The only prerequisite to use the dockerized Validate Tool is the availability of docker software in the host computer.
-At the time of writing this, the Validate Tool was tested with Docker version 20.10.12.
+To build an image, run:
 
-## 🏃 Steps to build the docker image of the Validate Tool
+    docker image build --build-arg tar_file=URL --tag [OWNER/]registry-api-service:TAG .
```
ramesh-maddegoda: Do we have to mention about the way to get this URL?

tloubrieu-jpl: Maybe, it would be safer. Is it a tar.gz ? You can give an example, others will easily extrapolate from the example to point to other versions.

Also the tag should be `validate`, not `registry-api-service`

jordanpadams: @ramesh-maddegoda @tloubrieu-jpl updated and works

jordanpadams: @ramesh-maddegoda @tloubrieu-jpl @nutjob4life 

do we need to support URLs or since this is part of the CICD mean we can just use local files?

Using `COPY` seems to fix the issue with the docker container not building, but then it cannot accept URLs.

The issue I was seeing using `ADD` is because for some reason, this also includes a hidden step where it decompresses the tar.gz into a folder. So we would need to add additional logic to figure all that out, or we could just assume local file path and be done with it.

Referring to comments below, the ADD [documentation](https://docs.docker.com/engine/reference/builder/#add) states that:
> If is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory. Resources from remote URLs are not decompressed.

nutjob4life: I think `ADD` has too much magic 😇

jordanpadams: @nutjob4life +1 agreed. the fact that it was doing that without being very clear that it was doing that is weird to me... will stick with local path for now.

PR #804, opened 2023-12-19
Improved report generation to be more consistent across output formats

### LINKED ISSUE #797: As a user, I want to have the same information available in the validation report no matter which style I choose
### Checked for duplicates

Yes - I already checked

### 🧑‍🔬 User Persona(s)

Anyone using validate with a `--report-style` other than "full".

### 💪 Motivation

So that the availability of information presented in the validate report isn't dependent on the selected output style.

### 📖 Additional Details

The `--report-style` or `-s` command-line option can be one of "full", "xml", or "json". The final portion of a "full" validate report (where referential integrity checks were performed) looks like:
```
Summary:

  0 error(s)
  X warning(s)

  Product Validation Summary:
    Y         product(s) passed
    0          product(s) failed
    0          product(s) skipped

  Referential Integrity Check Summary:
    Z          check(s) passed
    0          check(s) failed
    0          check(s) skipped

  Message Types:
    X          whatever message type

End of Report
```

Running validate on the same dataset but with `--report-style xml` will produce a report with a summary element at the end that looks like
```xml
<summary>
   <totalErrors>0</totalErrors>
   <totalWarnings>X</totalWarnings>
   <messageTypes>
      <messageType total="X">whatever message type</messageType>
   </messageTypes>
</summary>
```
(i.e., missing the breakdown by product validation and referential integrity check when compared to the former).

The summary portion of the JSON output (again under the same conditions except the report style) looks like:
```json
"summary": {
  "totalProducts": N,
  "totalErrors": 0,
  "totalWarnings": X,
  "messageTypes": [
    {
      "messageType": "whatever message type",
      "total": X
    }
  ]
}
```

This output _adds_ a very useful `"totalProducts"` key relative to the previous two, but omits the "full" report's information about product level validation and referential integrity.

So the content in the summary area for each output type is
| Report style                             | Full (plaintext) | XML | JSON |
|------------------------------------------|:----------------:|:---:|:----:|
| Total products                           |        :x:[^1]       | :x: | :heavy_check_mark: |
| Total errors                             | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Total warnings                           | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Product-level passes/fails/skips         | :heavy_check_mark: | :x: |  :x: |
| Referential integrity passes/fails/skips | :heavy_check_mark: | :x: |  :x: |
| Message type counts                      | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |

[^1]: But could be calculated by summing the product-level passes, failures, and skips

### Acceptance Criteria

**Given** 
**When I perform** 
**Then I expect** 

### ⚙️ Engineering Details

_No response_

al-niessner:
## 🗒️ Summary
Tried to keep as much of the original code as possible mainly for its integration throughout validate. Did change it to be more of a visitor pattern such that Report is responsible for the bulk of the content and format while the implementers of Report are responsible for serialization only. Not perfect to avoid large divergence from legacy, but much better.

## ⚙️ Test Data and/or Report
Unit tests should all pass otherwise JSON report is not similar enough.

## ♻️ Related Issues
Closes #797

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/report/FullReport.java
+++ b/src/main/java/gov/nasa/pds/validate/report/FullReport.java
@@ -75,7 +75,6 @@ public class FullReport extends Report {
   @Override
   protected void append(String title) {
     this.getWriter().println();
-    this.getWriter().println();
     this.getWriter().println(title);
   }
 
@@ -179,8 +178,6 @@ public class FullReport extends Report {
         break;
       case HEADER:
         this.getWriter().println();
-        this.getWriter().println("PDS Validate Tool Report");
-        this.getWriter().println();
         this.getWriter().println("Configuration:");
         for (Tuple configuration : this.configurations) {
           this.getWriter().print ("   ");
@@ -197,7 +194,7 @@ public class FullReport extends Report {
           this.getWriter().println(parameter.b);
         }
         this.getWriter().println();
-        this.append ("Product Level Validation Results");
+        this.getWriter().println();
         break;
       case LABEL:
         processMaps();
--- a/src/main/java/gov/nasa/pds/validate/report/FullReport.java
+++ b/src/main/java/gov/nasa/pds/validate/report/FullReport.java
@@ -35,7 +35,6 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import gov.nasa.pds.tools.label.ExceptionType;
 import gov.nasa.pds.tools.util.Utility;
 import gov.nasa.pds.tools.validate.ContentProblem;
 import gov.nasa.pds.tools.validate.ValidationProblem;
@@ -82,19 +81,7 @@ public class FullReport extends Report {
   protected void append(ValidationProblem problem) {
     if (this.currentStatus == Status.SKIP) {
       this.getWriter().print("      ");
-      String severity = "";
-      if (problem.getProblem().getSeverity() == ExceptionType.FATAL) {
-        severity = "FATAL_ERROR";
-      } else if (problem.getProblem().getSeverity() == ExceptionType.ERROR) {
-        severity = "ERROR";
-      } else if (problem.getProblem().getSeverity() == ExceptionType.WARNING) {
-        severity = "WARNING";
-      } else if (problem.getProblem().getSeverity() == ExceptionType.INFO) {
-        severity = "INFO";
-      } else if (problem.getProblem().getSeverity() == ExceptionType.DEBUG) {
-        severity = "DEBUG";
-      }
-      this.getWriter().print(severity);
+      this.getWriter().print(problem.getProblem().getSeverity().getName());
       this.getWriter().print("  ");
       this.getWriter().print("[" + problem.getProblem().getType().getKey() + "]");
       this.getWriter().print("   ");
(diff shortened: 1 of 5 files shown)


src/main/java/gov/nasa/pds/validate/report/FullReport.java
```diff
@@ -54,91 +52,157 @@
...
+        severity = "FATAL_ERROR";
+      } else if (problem.getProblem().getSeverity() == ExceptionType.ERROR) {
+        severity = "ERROR";
+      } else if (problem.getProblem().getSeverity() == ExceptionType.WARNING) {
+        severity = "WARNING";
+      } else if (problem.getProblem().getSeverity() == ExceptionType.INFO) {
+        severity = "INFO";
+      } else if (problem.getProblem().getSeverity() == ExceptionType.DEBUG) {
+        severity = "DEBUG";
+      }
```
nutjob4life: Should we have a final `else` in case the severity doesn't match any of the conditions? (Or default `severity` to `unknown` instead of empty string?)

nutjob4life (approved):
👍

src/main/java/gov/nasa/pds/validate/report/FullReport.java
al-niessner: To really fix it, change ExceptionType.toString() to do the right thing so that we can get rid of the if block. Can do and it may already be that way but this code does not use it and I just did a gut-n-paste from the old stuff since it "works".

al-niessner:
The json is not being reset well enough and the unit tests are breaking because of it. Working through it.

src/main/java/gov/nasa/pds/validate/report/FullReport.java
al-niessner: @nutjob4life 

You sure? I have the spare time. Do want to start a new big job but little clean up is good. Still fixing cucumber too.

al-niessner: @nutjob4life 

Yeah, my memory was right and getName() was already implemented correctly. Done.

al-niessner:
@jordanpadams @nutjob4life 

Did something change outside of this repo? If I use eclipse and JUnit these tests pass. If I do mvn build they pass. If I push here, they fail.

nutjob4life:
@al-niessner they don't work locally for me; when I run `mvn package` it fails during the testing phase. I get lots of `java.lang.IllegalStateException: Nesting problem`, just like GitHub Actions does.

al-niessner:
> @al-niessner they don't work locally for me; when I run `mvn package` it fails during the testing phase. I get lots of `java.lang.IllegalStateException: Nesting problem`, just like GitHub Actions does.

@nutjob4life 

Thanks for the sanity check. I found what I was doing wrong, then the bug, then pushed the fix.

al-niessner:
@jordanpadams @nutjob4life 

Well, I do not know what secrets I introduced typing code, but wow. Maybe return or this is a server password. Anyway, I tried the link in the details of failure:

https://nasa-ammos.github.io/slim/continuous-testing/starter-kits/#detect-secrets

and get a wonderful 404 with lots of nasa logos. Browsed around awhile and found it here:

https://nasa-ammos.github.io/slim/docs/guides/software-lifecycle/security/secrets-detection/

Makes no sense on what I am supposed to do since it is all about hooking it up and not about seeing what false positive it found and waiving it.

jordanpadams:
@nutjob4life @al-niessner The secrets detection has been failing since we introduced some updates a couple PRs ago. I have been trying to run the software, but it just hangs. Will take that up on a separate thread.

jordanpadams:
@al-niessner per the PR updates, how backwards incompatible, if at all, are the changes? I know we have some users that parse the JSON in their pipelines, so we want to do a major version bump if this will break those readers.

jordanpadams:
@al-niessner also, the secrets issue has been fixed on main, so that failure will go away once it is merged.

al-niessner:
@jordanpadams 

Is there a problem with pds site? All the XSD files are coming up 404.

jordanpadams:
@al-niessner it is just the XSD/SCH associated with PSA. Their site was upgraded and those schemas were missing. I have notified their team, who will hopefully get them back online ASAP.

jordanpadams:
Ran the tests successfully locally (commenting out ESA-related data). Backwards compatibility confirmed, for the most part, for the JSON output, which appears to be used more than others for reporting automation.

PR #818, opened 2024-01-24
Fix bug introduced by #308 when converting URLs to local paths on Windows

### LINKED ISSUE #809: Validate fails to read files on Windows systems after Internal Reference check updates (#308)
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

For a CSV containing multiple tables, validate 3.1.1 reports the following error on a table that we think is properly defined in the label:
```
ERROR  [error.label.invalid_object_definition]   line 2: Invalid object definition for object #2.  The previously defined object ends at byte 354958. Offset defined in label: 25
ERROR  [error.label.invalid_object_definition]   line 3: Invalid object definition for object #3.  The previously defined object ends at byte 354958. Offset defined in label: 7751  
Note: The PDS4Viewer opens this file with no issues.
```

We tried to check with validate 3.4.1 to see if this issue has been fixed already, but we get another error:
```
FAIL: file:/X:/staff-folders/ward/clipper/RSS001E01_2031066T0241_EURGRVL20XXXXXXGSFC_COV010.xml
      ERROR  [error.validation.internal_error]   Uncaught exception while validating: null
```

### 🕵️ Expected behavior

We think the CSV and XML are valid and validate shouldn't be reporting any errors.

### 📜 To Reproduce
```
validate RSS001E01_2031066T0241_EURGRVL20XXXXXXGSFC_COV010.xml -R pds4.label -r report.txt
(note: you will get errors due to the absence of the Clipper Mission dictionary which isn't online yet. I can send a copy if helpful.)
```

### 🖥 Environment Info

- Version of this software: v3.1.1, v3.4.1
- Operating System: Windows

Validate v3.1.1 and  Validate v3.4.1

[validate_bug.zip](https://github.com/NASA-PDS/validate/files/13923643/validate_bug.zip)

### 🦄 Related requirements

🦄 #308

User also indicated seeing this on the discussion board: https://github.com/NASA-PDS/validate/discussions/811

### ⚙️ Engineering Details

After some investigation and testing on Windows, we started getting these errors on Windows when trying to resolve a path:
```
Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <:> at index 2: 
...
```

which can be traced to https://github.com/NASA-PDS/validate/blob/main/src/main/java/gov/nasa/pds/tools/validate/CrossLabelFileAreaReferenceChecker.java#L19

Per https://stackoverflow.com/questions/43972777/java-nio-file-invalidpathexception-illegal-char-at-index-2, on Windows, we need to translate the URL to a URI vs. a path.

### I&T
TestRail Test ID: T8681197

jordanpadams:
## 🗒️ Summary
When using Path.of() we need to pass a URI on a Windows machine, not the Path.

Resolves #809

## ⚙️ Test Data and/or Report
See regression tests to verify this doesn't break other tests.

## ♻️ Related Issues

Resolves #809

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/CrossLabelFileAreaReferenceChecker.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/CrossLabelFileAreaReferenceChecker.java
@@ -1,6 +1,7 @@
 package gov.nasa.pds.tools.validate;
 
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -14,9 +15,9 @@ import gov.nasa.pds.tools.util.LabelUtil;
 
 public class CrossLabelFileAreaReferenceChecker {
   final private static HashMap<String,List<String>> knownRefs = new HashMap<String,List<String>>();
-  private static String resolve (String name, ValidationTarget target) {
+  private static String resolve (String name, ValidationTarget target) throws URISyntaxException {
     if (!name.startsWith ("/")) {
-      name = Path.of(target.getUrl().getPath()).getParent().resolve(name).toString();
+      name = Path.of(target.getUrl().toURI()).getParent().resolve(name).toString();
     }
     return name;
   }
@@ -30,8 +31,9 @@ public class CrossLabelFileAreaReferenceChecker {
    * @throws IOException
    * @throws ParserConfigurationException
    * @throws SAXException
+   * @throws URISyntaxException
    */
-  public static boolean add (String name, ValidationTarget target) throws IOException, ParserConfigurationException, SAXException {
+  public static boolean add (String name, ValidationTarget target) throws IOException, ParserConfigurationException, SAXException, URISyntaxException {
     boolean success = false;
     DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
     Document xml = dbf.newDocumentBuilder().parse(target.getUrl().openStream());
@@ -49,10 +51,10 @@ public class CrossLabelFileAreaReferenceChecker {
     }
     return success;
   }
-  public static String getOtherId (String name, ValidationTarget target) {
+  public static String getOtherId (String name, ValidationTarget target) throws URISyntaxException {
     return knownRefs.get(resolve(name, target)).get(0);
   }
-  public static String getOtherFilename (String name, ValidationTarget target) {
+  public static String getOtherFilename (String name, ValidationTarget target) throws URISyntaxException {
     return knownRefs.get(resolve(name, target)).get(1);
   }
   public static void reset() {


jordanpadams:
@al-niessner can you poke at this when you can? This is breaking validate on Windows.

al-niessner:
> @al-niessner can you poke at this when you can? This is breaking validate on Windows.

@jordanpadams 

Are you saying that getURI() is not the solution over getPath() so it still does not work? If so, can we undo these changes. I dislike changes that do not fix anything but add complication (new exception handling for no reason).

al-niessner:
@jordanpadams 

Just read the ticket and error. The problem is not getPath(). It is the URL is malformed somehow according the [windoze experts](http://help.collab.net/index.jsp?topic=/faq/winletterdrive.html). Also [wikipedia](https://en.wikipedia.org/wiki/File_URI_scheme)

Let me see if I can round up a windoze box and test a little bit.

al-niessner:
@jordanpadams 

Nope. We are both wrong it seems. JDK 17? I have eclipse running jdk 17 running a simple test and they both work. You can run the test on your windows to see how it does:

Code:
```
$ cat ~/Scratch/UrlOnWIn.java 
package test;

import java.net.MalformedURLException;
import java.net.URL;

public class UrlOnWIn {
	public static void main(String[] args) {
		try {
			URL a = new URL("file:/a:/foo/bar");
			URL b = new URL("file:///a:/foo/bar");
			
			try { a.getPath(); System.out.println ("a worked"); }
			catch (Exception e) { System.out.println ("a failed"); }
			try { b.getPath(); System.out.println ("b worked"); }
			catch (Exception e) { System.out.println ("b failed"); }
		} catch (MalformedURLException e) {
			e.printStackTrace();
		}
	}
}
```

Result:
```
a worked
b worked
```

al-niessner:
@jordanpadams 

LOL - experts were right. The failure is from the Path.of() not in the URL.getPath(). Gads. Anyway, update the test and we are clearly forming the URL incorrectly when we make on windows. 

Code:
```
package test;

import java.net.MalformedURLException;
import java.net.URL;
import java.nio.file.Path;

public class UrlOnWIn {
	public static void main(String[] args) {
		try {
			URL a = new URL("file:/a:/foo/bar");
			URL b = new URL("file:///a:/foo/bar");
			
			try { a.getPath(); System.out.println ("a worked"); System.out.println (Path.of(a.getPath()));}
			catch (Exception e) { System.out.println ("a failed"); }
			try { b.getPath(); System.out.println ("b worked"); }
			catch (Exception e) { System.out.println ("b failed"); System.out.println (Path.of(a.getPath()));}
		} catch (MalformedURLException e) {
			e.printStackTrace();
		}
	}
}
```

Result:
```
a worked
a failed
b worked
```

So, I can chase this back if you like but other than doing simple tests in wine, I cannot really do much.

al-niessner:
@jordanpadams 

Flipping through browser tabs and noticed test is not right. Give me a moment. Have to start wine again.

al-niessner:
@jordanpadams 

Okay, test is now more pedantic. It is a mess because it is so hard for me to use windoze. The answer is, unfortunately, that we need the URI **and** the URL needs to be fixed. The test results show that if you need/want the x: to become \\x in Path, then you need the right number of flipping slashes in the URL. I expanded the testing but it is still hard to read the output but the hint is bb is right and all others are wrong. Back when we create the URL probably should set the protocol rather than prefix with what seems to work.

Code:
```
package test;

import java.net.MalformedURLException;
import java.net.URI;
import java.net.URISyntaxException;
import java.net.URL;
import java.nio.file.Path;

public class UrlOnWIn {
	public static void main(String[] args) {
		try {
			URL a = new URL("file:/a:/foo/bar");
			URL b = new URL("file://b:/foo/bar");
			URL c = new URL("file://c:/foo/bar");
			
			try { System.out.println ("URL: " + a.getPath()); System.out.println ("Path: " + Path.of(a.getPath())); }
			catch (Exception e) { System.out.println ("a failed"); }
			try { System.out.println ("URL: " + b.getPath()); System.out.println ("Path: " + Path.of(b.getPath())); }
			catch (Exception e) { System.out.println ("b failed"); }
			try { System.out.println ("URL: " + c.getPath()); System.out.println ("Path: " + Path.of(c.getPath())); }
			catch (Exception e) { System.out.println ("c failed"); }
			
			try { System.out.println ("URL: " + a.getPath()); System.out.println ("Path: " + Path.of(a.toURI())); }
			catch (Exception e) { System.out.println ("a failed"); }
			try { System.out.println ("URL: " + b.getPath()); System.out.println ("Path: " + Path.of(b.toURI())); }
			catch (Exception e) { System.out.println ("b failed"); }
			try { System.out.println ("URL: " + c.getPath()); System.out.println ("Path: " + Path.of(c.toURI())); }
			catch (Exception e) { System.out.println ("c failed"); }
			
			URI aa = new URI("file:/aa:/foo/bar");
			URI bb = new URI("file://bb:/foo/bar");
			URI cc = new URI("file:///cc:/foo/bar");
			try { System.out.println ("URI: " + aa.getPath()); System.out.println ("Path: " + Path.of(aa)); }
			catch (Exception e) { System.out.println ("aa failed"); }
			try { System.out.println ("URI: " + bb.getPath()); System.out.println ("Path: " + Path.of(bb)); }
			catch (Exception e) { System.out.println ("bb failed"); }
			try { System.out.println ("URI: " + cc.getPath()); System.out.println ("Path: " + Path.of(cc)); }
			catch (Exception e) { System.out.println ("cc failed"); }			
		} catch (MalformedURLException e) {
			e.printStackTrace();
		} catch (URISyntaxException e1) {
			e1.printStackTrace();
		}
	}
}
```

Output:
```
URL: /a:/foo/bar
a failed
URL: /foo/bar
Path: \foo\bar
URL: /foo/bar
Path: \foo\bar
URL: /a:/foo/bar
Path: a:\foo\bar
URL: /foo/bar
Path: \\b\foo\bar
URL: /foo/bar
Path: \\c\foo\bar
URI: /aa:/foo/bar
aa failed
URI: /foo/bar
Path: \\bb\foo\bar
URI: /cc:/foo/bar
cc failed
```

jordanpadams:
@al-niessner just did a quick test of this and it looks like this is a weird thing with the URL vs. URI toStrings?

```
[main] DEBUG gov.nasa.pds.validate.ValidateLauncher - Path.of(t).toUri(): file:///absolute/path/to/ff_char.xml
[main] DEBUG gov.nasa.pds.validate.ValidateLauncher - Path.of(t).toUri().toURL():  file:/absolute/path/to/ff_char.xml
```

So I'm not sure what to do here? Do we need to fix the URL? Per this, it sounds like it is a difference in the `toString()`? https://stackoverflow.com/questions/6098472/pass-a-local-file-in-to-url-in-java

jordanpadams:
@al-niessner nevermind, I think I see the issue. Once convert to URL, it loses the appropriate path information.

```
        url = Path.of(t).toUri().toURL();
        LOG.debug("URI: {}", url.toURI());
        LOG.debug("URL: {}", url);
```

```
[main] DEBUG gov.nasa.pds.validate.ValidateLauncher - URI: file:/path/github681/ff_char.xml
[main] DEBUG gov.nasa.pds.validate.ValidateLauncher - URL: file:/path/github681/ff_char.xml
```

jordanpadams:
@al-niessner I have been testing this like crazy and cannot figure out a workaround using URL objects. I think we will need to completely refactor out URLs to URIs throughout the code.

That being said, this code appears to now run successfully on windows. Are you saying this does not/should not work? Or that this is bound to break elsewhere in the code if we don't fix this path issue?

jordanpadams:
Merging this PR for now since it does fix something. But per comment above, may not fix everything.

msbentley:
Hi @jordanpadams when will a version with this fix be released?

Issue #827, opened 2024-02-16
As a user, I want to have the parameter names used in 3.5.0 reports to be the same as they were in ≤3.4.1


cgobat:
### Checked for duplicates

Yes - I've already checked

### 🧑‍🔬 User Persona(s)

All users, especially those who do any kind of report parsing programmatically.

### 💪 Motivation

To minimize impact of recent updates to validate report structures on existing pipelines.

### 📖 Additional Details

This is a followup to my conversation with @al-niessner on #797. As requested in [this comment](https://github.com/NASA-PDS/validate/issues/797#issuecomment-1930673176), I'm opening this issue to serve as an "official" request to use the following set of keywords in the parameters section of the validation reports.

| Existing (≤3.4.1)        | Updated (3.5.0-SNAPSHOT)  | Requested (for 3.5.0 release)  |
|:-------------------------|:--------------------------|:-----------------------------|
| `targets`                | `targets`                 |  *no change*                 |
| `ruleType`               | `ruleType`                |  *no change*                 |
| `severityLevel`          | `level`                   | `severityLevel`              |
| `recurseDirectories`     | `recurse`                 | `recurseDirectories`         |
| `fileFiltersUsed`        | `regex`                   | `fileFiltersUsed`            |
| `dataContentValidation`  | `contentValidation`       | `dataContentValidation`      |
| `productLevelValidation` | `productLevelValidation`  |  *no change*                 |
| `allowUnlabeledFiles`    | `allowUnlabeledFiles`     |  *no change*                 |
| `maxErrors`              | `maxErrs`                 | `maxErrors`                  |
| `registeredContextsFile` | `regProds`                | `registeredContextsFile`     |
| `nonRegisteredContextsFile` | `nonRegProds` | `nonRegisteredContextsFile` |
| `dataEveryN` | `everyN` | `dataEveryN` |
| `dataSpotCheck` | `spotCk` | `dataSpotCheck` |
| `completeDescriptions` | `descriptions` | `completeDescriptions` |
| `checksumManifestFile` | `manifestCksum` | `checksumManifestFile` |
| `manifestFileBasePath` | `manifestPath` | `manifestFileBasePath` |
| `pDFErrorDirectory` | `pdfErrDir` | `pdfErrorDirectory`[^1] |
| `userSpecifiedSchemas` | `schemas` | `userSpecifiedSchemas` |
| `userSpecifiedCatalogs` | `catalogs` | `userSpecifiedCatalogs` |
[...]

PR #835, opened 2024-02-21
Fix plurality in JSON report header to make it consistent with other styles

cgobat:
## 🗒️ Summary
Removes the `s` from `"configurations"` so that the JSON report header keys match the tags in the XML style (see [line 150 of XmlReport.java](https://github.com/NASA-PDS/validate/blob/main/src/main/java/gov/nasa/pds/validate/report/XmlReport.java#L150)) and the section titles in the "full" report (see [line 168 of FullReport.java](https://github.com/NASA-PDS/validate/blob/main/src/main/java/gov/nasa/pds/validate/report/FullReport.java#L168)).

## ⚙️ Test Data and/or Report
Given the extremely simple nature of this PR (see [Files changed](./835/files)), I did not bother setting up a local testing environment. Hopefully that is okay in this case, but let me know if it's needed.

## ♻️ Related Issues

https://github.com/NASA-PDS/validate/issues/827#issuecomment-1949434226

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/report/JSONReport.java
+++ b/src/main/java/gov/nasa/pds/validate/report/JSONReport.java
@@ -243,7 +243,7 @@ public class JSONReport extends Report {
           this.msgs.clear();
           break;
         case HEADER:
-          this.append ("configurations", this.configs);
+          this.append ("configuration", this.configs);
           this.append ("parameters", this.params);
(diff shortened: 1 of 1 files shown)


nutjob4life (approved):
This looks safe!

al-niessner (approved):
Good fix.

PR #734, opened 2023-10-19
Upgrade net.sf.saxon:Saxon-HE from 9.9.1-7 to 12.3

al-niessner:
## 🗒️ Summary
The new SAXON libraries are more pedantic about processing XML files. It now generates warning and errors that cannot be squashed - well I could not find the wire to cut to the blinking red light. Instead, changed validate so that it would not ask SAXON to parse files that do not have the XML extension given by the user or defined by default.

## ⚙️ Test Data and/or Report
All automated unit tests should pass

## ♻️ Related Issues
Closes #707
Closes #726

Issue #597, opened 2023-02-17
Validate does not maintain history of other versions it comes across as it traverses directories causing erroneous WARNING messages


jordanpadams:
### Checked for duplicates

Yes - I have checked

### 🐛 Describe the bug

Per https://github.com/NASA-PDS/operations/issues/322#issuecomment-1377978726 and the validate report denoted in https://github.com/NASA-PDS/operations/issues/322, validate is throwing erroneous WARNING messages for past versions of bundles/collections in the directory used as the target.

### 🕵️ Expected behavior

Validate would maintain an understanding of past versions of bundles and the past versions of collections they refer to, and simple SKIP validation of those products without throwing a WARNING.

### 📜 To Reproduce

1. download either of the test data sets
2. run `validate -R pds4.bundle -t /path/to/naif/dataset`
```
...
  PASS: file:/Users/jpadams/test/naif3/spice_kernels/collection_spice_kernels_inventory_v001.csv
      WARNING  [warning.file.not_referenced_in_label]   File is not referenced by any label
        3 integrity check(s) completed

  PASS: file:/Users/jpadams/test/naif3/spice_kernels/collection_spice_kernels_inventory_v002.csv
      WARNING  [warning.file.not_referenced_in_label]   File is not referenced by any label
        4 integrity check(s) completed
...
Summary:

  0 error(s)
  2 warning(s)

  Product Validation Summary:
    17         product(s) passed
    0          product(s) failed
    4          product(s) skipped

  Referential Integrity Check Summary:
    19         check(s) passed
    0          check(s) failed
    0          check(s) skipped

  Message Types:
    2            warning.file.not_referenced_in_label
```

### 🖥 Environment Info

- Version of this software: 3.1.1
- Operating System: MacOSX

### 📚 Version of Software Used

v3.1.1

### 🩺 Test Data / Additional context

https://naif.jpl.nasa.gov/pub/naif/pds/pds4/mars2020/mars2020_spice/

or for a smaller data set:

https://pds.nasa.gov/data/pds4/test-data/custom-datasets/naif3/

### ⚙️ Engineering Details

Looks like regression was introduced prior to v3.1.1. Do not see this in v3.0.3

jordanpadams:
@al-niessner wrap up your current ticket, and then let's loop back to this and another ticket I am about to create that have introduced a couple regressions

al-niessner:
@jordanpadams @nutjob4life @tloubrieu-jpl 

[...]

PR #836, opened 2024-02-22
Fix regression where unlabeled files are not being identified

### LINKED ISSUE #822: Check for unlabeled files no longer works
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I run validate it no longer flags unlabled-files, even if the --allow-unlabled-files option is not called.

### 🕵️ Expected behavior

I expect that when I run validate that it will warn me when there are files that do not have labels, as prior versions did.

### 📜 To Reproduce

1. Grab a bundle or collection and ensure there is an unlabeled file in the directory.
2. Run validate.  Example:  validate -R pds4.bundle -t [point to bundle file OR bundle directory]
3.
...

### 🖥 Environment Info

- Version of this software tested: 3.3.3 and 3.4.1 (note it works in version 3.3.0 and prior)
- Operating Systems: MacOSX 12.6.8 and Red Hat Enterprise Linux 9.3
...

gov.nasa.pds:validate
Version 3.4.1
Release Date: 2023-12-14 00:22:59

All rights reserved.

gov.nasa.pds:validate
Version 3.3.3
Release Date: 2023-11-20 23:43:57

All rights reserved.

_No response_

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

al-niessner:
## 🗒️ Summary
Put some code back in that got lost during the saxon upgrade #734. Probably removed in an attempt to make something work but it was not the cause and then left it out since none of the regression tests failed.

## ⚙️ Test Data and/or Report
See automated tests of this PR as they should pass

## ♻️ Related Issues
Closes #822


al-niessner:
@jordanpadams 

Which 7 files do we expect to be unlabeled?

jordanpadams:
@al-niessner not positive if these are referenced in a label or not, but I don't have time to look through of the labels to verify. Here is my guess (it may only be 6...):
* src/test/resources/github6/invalid//data_spectra/000DANGLING_FILE_WITH_EXTENSION.txt
* src/test/resources/github6/invalid//data_spectra/000DANGLING_FILE_WITHOUT_EXTENSION
* src/test/resources/github6/invalid//core
* src/test/resources/github6/invalid//file_aux_should_be_ok.tab
* src/test/resources/github6/invalid//core.vmac
* src/test/resources/github6/invalid//file_with_bad_char_#_.txt

al-niessner:
> @al-niessner not positive if these are referenced in a label or not, but I don't have time to look through of the labels to verify. Here is my guess (it may only be 6...):
> 
> * src/test/resources/github6/invalid//data_spectra/000DANGLING_FILE_WITH_EXTENSION.txt
> * src/test/resources/github6/invalid//data_spectra/000DANGLING_FILE_WITHOUT_EXTENSION
> * src/test/resources/github6/invalid//core
> * src/test/resources/github6/invalid//file_aux_should_be_ok.tab
> * src/test/resources/github6/invalid//core.vmac
> * src/test/resources/github6/invalid//file_with_bad_char_#_.txt

@jordanpadams I am going to move away from this test to #597 and PR #615 because it is much cleaner and potentially related. Long post coming in near future saying no, but still cleaner for testing.

al-niessner:
The code responsible for throwing this error is in one place and one place only:

https://github.com/NASA-PDS/validate/blob/918da20f5641443a8ceb146d97f327db2b1e0447/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/FindUnreferencedFiles.java#L86-L110

Blames says that the only change was a year ago in #615. It is possible that we now have false negatives but need to be very careful. One, it is looking only for targets (labels) that it found but are not used by aggregates. It will not find any random extra file because of the loop:

https://github.com/NASA-PDS/validate/blob/918da20f5641443a8ceb146d97f327db2b1e0447/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/FindUnreferencedFiles.java#L97

The loop function in question is here:

https://github.com/NASA-PDS/validate/blame/918da20f5641443a8ceb146d97f327db2b1e0447/src/main/java/gov/nasa/pds/tools/validate/InMemoryRegistrar.java#L165

and has not changed significantly in 5 years (did code reformat 2 years ago). My next guess is that some code block that used to use that problem type was axed. Therefore I rolled back to v3.3.0 which the user reported as working. Still just the one location (FindUnreferencedFiles). Interestingly, v3.3.0 contains #615.

It means that either it never detected any random file or what we stuffed in as a target could have been regular files if not referenced by a label. Preparing an approach for this testing but it will not be simple.

src/main/java/gov/nasa/pds/tools/validate/rule/RegisterTargets.java
```diff
@@ -48,6 +49,9 @@ public void registerTargets() {
       try {
         Crawler crawler = getContext().getCrawler();
         WildcardOSFilter fileFilter = getContext().getFileFilters();
+        if (!"PDS4 Directory".equalsIgnoreCase(getContext().getRule().getCaption())) {
```
al-niessner, line 52: This got removed in #734. Not sure why and did not leave myself a good comment. Without it, the filter is looking for just XML files.

nutjob4life (approved):

jordanpadams (approved):

Issue #844, opened 2024-02-27
Exception when trying to handle `missing_constant` == `Infinity`


jordanpadams:
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I tried validating a product with `missing_constant` == `Infinity`, I get a FATAL_ERROR.

Upon further investigation, it is throwing a `java.lang.NumberFormatException`

### 🕵️ Expected behavior

I expected validate to complete successfully.

### 📜 To Reproduce

```
$ validate-3.5.0-SNAPSHOT/bin/validate -t ~/Downloads/MDIS_RTM_N01_010599_7521824_1.xml

java.lang.NumberFormatException: For input string: "Infinity"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:668)
	at java.base/java.math.BigInteger.<init>(BigInteger.java:538)
	at gov.nasa.pds.tools.validate.content.SpecialConstantBitPatternTransforms.asBigInt(SpecialConstantBitPatternTransforms.java:25)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.sameContent(ArrayContentValidator.java:352)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.isSpecialConstant(ArrayContentValidator.java:372)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.validatePosition(ArrayContentValidator.java:295)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.process(ArrayContentValidator.java:161)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.process(ArrayContentValidator.java:152)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.process(ArrayContentValidator.java:152)
	at gov.nasa.pds.tools.validate.content.array.ArrayContentValidator.validate(ArrayContentValidator.java:126)
	at gov.nasa.pds.tools.validate.rule.pds4.ArrayValidator.validateDataObjectContents(ArrayValidator.java:104)
	at gov.nasa.pds.tools.validate.rule.pds4.ArrayValidator.validate(ArrayValidator.java:59)
	at gov.nasa.pds.tools.validate.rule.pds4.DataDefinitionAndContentValidationRule.validate(DataDefinitionAndContentValidationRule.java:127)
[...]

PR #889, opened 2024-05-03
Catch NumberFormatException when values in data do not match spec in label

al-niessner:
## 🗒️ Summary
Log as a validation error rather than fail from an exception.

## ⚙️ Test Data and/or Report
Automated unit/regression tests below pass

## ♻️ Related Issues
Temporary patch to #844 that allows validation to continue processing but does not fix or solve the actual problem. Waiting for SWG and/or DDWG to reach a decision for a solution.


nutjob4life (approved):
👍

jordanpadams (approved):

PR #914, opened 2024-05-30
Update duplicate file check to only check for Product Observational

### LINKED ISSUE #905: Validate 3.5.0-snapshot gives error when encountering Document objects/files used in other Document products.
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

Two versions of a Document include one ASCII text file and seven JPEG figure files; the ASCII files have a different file names, the JPEGs have the same file names.  Validate throws an error upon encountering the JPEG file names because they have been used previously.

### 🕵️ Expected behavior

Previous versions of Validate have not flagged this situation as an error.  SR 2A.3 (paragraph 2) prohibits use of observational data files in more than one product.  The JPEGs (and the parent Document) are not observational data.  There should be no error.
I'm guessing that Validate has recently been upgraded to check for multiple uses of data objects/files.  Its search needs to be narrowed to reuse of data objects containing observational data.

### 📜 To Reproduce

1. I have not done this; but I think it should give the desired result (contact guinness@wustl.edu for more info)
2. Unzip and untag the attached bug?.tar.gz files
3. Create a Collection using the two products
4. Run Validate on the Collection

### 🖥 Environment Info

- I don't know (contact guinness@wustl.edu)

validate 3.5.0-snapshot

[bug1.tar.gz](https://github.com/NASA-PDS/validate/files/15396090/bug1.tar.gz)
[bug2.tar.gz](https://github.com/NASA-PDS/validate/files/15396091/bug2.tar.gz)

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

### 🎉 Integration & Test

_No response_

al-niessner:
## 🗒️ Summary
Look for the word observational in the file area and any of its ancestors.

## ⚙️ Test Data and/or Report
Automated unit/regression tests below should pass

## ♻️ Related Issues
Fixes #905


src/main/java/gov/nasa/pds/tools/validate/CrossLabelFileAreaReferenceChecker.java
```diff
@@ -33,20 +34,24 @@ private static String resolve (String name, ValidationTarget target) throws URIS
...
+  public static boolean add (String name, ValidationTarget target, boolean isObs)
+      throws IOException, ParserConfigurationException, SAXException, URISyntaxException {
     boolean success = false;
     DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
     Document xml = dbf.newDocumentBuilder().parse(target.getUrl().openStream());
-    DOMSource domSource = new DOMSource(xml);    
+    DOMSource domSource = new DOMSource(xml);
+    String full_name = resolve(name, target);
+    isObservational.put(full_name,
+        (isObservational.containsKey(full_name) ? isObservational.get(full_name) : false) | isObs);
```
nutjob4life: Using the bitwise OR `|` here instead of the logical OR `||` means Java will always evaluate the right-hand side. Maybe this is a premature optimization, but perhaps it could be rewritten with `isObs || (isObservational…` and let `isObs` being `true` skip the `containsKey`/`get` call completely?

nutjob4life (approved):
Approved, but see my question re `|` vs `||` 😁

src/main/java/gov/nasa/pds/tools/validate/CrossLabelFileAreaReferenceChecker.java
al-niessner: Yes, that would be a CPU time improvement. Changed.

PR #922, opened 2024-06-07
Allow 64 bits to span more than 8 bytes (think shifted relative to byte boundaries)

### LINKED ISSUE #919: Validate throws an error when UnsignedBitString has 61 bits
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I wrote a label which describes an entire file as a series of bit strings, everything validated when strings were 60 bits or less.  When I tried to define a bit string with 61 bits, I got this error:
"Bit field spans bytes that are wider than a long (9 > 8)"
SR 5C.4 says bit strings are limited to 64 bits; it says nothing about how they span byte boundaries.

### 🕵️ Expected behavior

I expected no errors for bit strings up to 64 bits.

### 📜 To Reproduce

[validate_bit_string.tar.gz](https://github.com/user-attachments/files/15542805/validate_bit_string.tar.gz)

2. Run validate on the *.xml file; note that it fails.
3. In *.xml file, bit field 55, change the value of stop_bit_location to 1392
4. Rerun validate; it should pass.
...

### 🖥 Environment Info

- Version of this software: validate 3.4.1
- Operating System: Mac OS 14.5
...

validate 3.4.1

see above

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

### 🎉 Integration & Test

_No response_

al-niessner:
## 🗒️ Summary
Add a regression check only,

## ⚙️ Test Data and/or Report
Automated unit/regression tests below should pass

## ♻️ Related Issues
Closes #919


nutjob4life (approved):
👍

al-niessner:
@jordanpadams here is your unit test for regression

PR #930, opened 2024-06-10
Improve error handling for invalid config file options

### LINKED ISSUE #923: Configuration file parser does not reject incorrect options
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

I received a report from a data provider about unusual errors that they were encountering with validate. One of the issues that showed up in the report indicated that their schemas were not being loaded. The configuration file that they provided indicated that they used the option "validate_schema" instead of "validate.schema", so the schemas were not being loaded correctly.

### 🕵️ Expected behavior

I expected validate to fail when it encounters configuration options that it cannot handle.

### 📜 To Reproduce

create a configuration file that uses validate_schema instead of validate.schema and save it to config.txt
run `validate -c config.txt -t target`
Look for the " User Specified Schemas " section at the top of the report, which will be missing.

### 🖥 Environment Info

- Version of this software validate3.5.1
- Operating System: MacOSX with openjdk 21.0.3
...

validate 3.5.1

_No response_

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

### 🎉 Integration & Test

_No response_

al-niessner:
## 🗒️ Summary
Brief summary of changes if not sufficiently described by commit messages.

## ⚙️ Test Data and/or Report
Not automated test. New output when there is an unknown keyword in configuration file is:
```
validate -c config.txt -t something
org.apache.commons.configuration.ConfigurationException: Unrecognized keyword(s) in config file: bad.keyword
```

## ♻️ Related Issues
Closes #923

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/commandline/options/ConfigKey.java
+++ b/src/main/java/gov/nasa/pds/validate/commandline/options/ConfigKey.java
@@ -30,6 +30,11 @@
 
 package gov.nasa.pds.validate.commandline.options;
 
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * An interface that contains the valid property keys for the Validate Tool configuration file.
  *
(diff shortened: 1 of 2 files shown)


al-niessner:
@jordanpadams @nutjob4life 

Can safely ignore JAXA error in auto tests. Just simple code change here.

src/main/java/gov/nasa/pds/validate/commandline/options/ConfigKey.java
```diff
@@ -159,4 +164,26 @@ public class ConfigKey {
...
+  public static final Set<String> ALL_KEYWORDS;
+  static {
+    ALL_KEYWORDS = new HashSet<String>();
+    for (Field field : ConfigKey.class.getDeclaredFields()) {
+      if (Modifier.isStatic(field.getModifiers()) && field.getType().equals(String.class)) {
+        try {
+          ALL_KEYWORDS.add(field.get(null).toString());
+        } catch (IllegalArgumentException | IllegalAccessException e) {
+          // can we every get here??
+          e.printStackTrace();
```
nutjob4life: `every` → `even`

However, if getting "here" is an unexpected situation, maybe aborting is the right approach rather than just printing the stack trace and soldering on? Maybe throw a new `IllegalStateException`. Since that's a `RuntimeException` you don't have to declare it.

Thinking further: `IllegalArgumentException` is itself a `RuntimeException` so you don't even have to catch it. Only `IllegalAccessException` needs to be handled—and maybe aborting if it occurs is the right thing? I dunno 🤷

nutjob4life (approved):
Not requesting changes so much as requesting "reflection" on my comment above 😁

src/main/java/gov/nasa/pds/validate/commandline/options/ConfigKey.java
al-niessner: Excellent suggestion. Done.

jordanpadams:
Branch testing re-run successful: https://github.com/NASA-PDS/validate/actions/runs/9487729408

jordanpadams (approved):

PR #941, opened 2024-06-18
Specify specific maven project.version in the validate execution scripts

al-niessner:
## 🗒️ Summary
Updated scripts so that maven filtering will do substitutions during maven build

## ⚙️ Test Data and/or Report
Outside of java and regression tests.

validate now looks like this after mvn build:
```
VALIDATE_JAR=${LIB_DIR}/validate-3.6.0-SNAPSHOT.jar
```

### DIFF
--- a/src/main/assembly/tar-assembly.xml
+++ b/src/main/assembly/tar-assembly.xml
@@ -56,7 +56,7 @@ POSSIBILITY OF SUCH DAMAGE.
       </excludes>
     </fileSet>
     <fileSet>
-      <directory>src/main/resources/bin</directory>
+      <directory>target/classes/bin</directory>
       <outputDirectory>bin</outputDirectory>
       <includes>
         <include>validate</include>
@@ -69,7 +69,7 @@ POSSIBILITY OF SUCH DAMAGE.
       <lineEnding>keep</lineEnding>
     </fileSet>
     <fileSet>
-      <directory>src/main/resources/bin</directory>
+      <directory>target/classes/bin</directory>
       <outputDirectory>bin</outputDirectory>
       <includes>
         <include>logging.properties</include>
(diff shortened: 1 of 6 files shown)


nutjob4life (approved):
Thanks for facing the windows 👍

jordanpadams (changes requested):
@al-niessner this doesn't seem to work for me? I ran `mvn package` and unpacked the package, and still have this in the script.

```
# Finds the jar file in LIB_DIR and sets it to VALIDATE_JAR.
VALIDATE_JAR=${LIB_DIR}/validate-@project.version@.jar
```

al-niessner:
@jordanpadams 

Oh, that is nuts. I did not look in the artifact I saw it change in target/... and figured it would package it. Let me double check it is fully working for me.

jordanpadams:
@al-niessner probably have to put the setting to resolve the variables in the assembly.xml

al-niessner:
@jordanpadams assembly.xml? Just did find in validate and no such file was found.

al-niessner:
@jordanpadams oh, found them.

al-niessner:
@jordanpadams the assemblies get them from src not target which explains why they are not filtered as the pom says...

al-niessner:
@jordanpadams 

Try again. I fixed the assemblies. I did not change the context json but if you are wondering why the POM is not updating it like you want, well it is because you are reading the source not the target.

jordanpadams:
@al-niessner the tar `validate` script still contains:

```
# Finds the jar file in LIB_DIR and sets it to VALIDATE_JAR.
VALIDATE_JAR=${LIB_DIR}/validate-@project.version@.jar

```

jordanpadams (approved):
Nevermind. User error.

PR #980, opened 2024-08-23
Performance improvements for referential integrity checking

### LINKED ISSUE #969: Slow performance with all content and product validation turned off
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I ran validate on a 2M product bundle with everything disabled expected referential integrity checks, it still takes 2 weeks.

Products: 2013873
Completed in 13.5 days / 324.6 hours

### 🕵️ Expected behavior

I expected the run to take less time.

### 📜 To Reproduce

1. get a large bundle.

2.

Here's the command:

```
/path/validate-3.5.1/bin/validate -target
/path/msam2/annex_ehlmann_caltech_msl_msam2 --report-file bundle.valrpt \
               -R pds4.bundle \
               --skip-content-validation --skip-product-validation
```

### 🖥 Environment Info

Linux OS
m7l.2xlarge EC2 instance

v3.5.1

_No response_

### 🦄 Related requirements

Tightly coupled with #931 

### ⚙️ Engineering Details

Per initial analysis by @al-niessner, [code block here](https://github.com/NASA-PDS/validate/blob/080b7c8862b544f7997ab7582fe7d9193a33a74a/src/main/java/gov/nasa/pds/tools/validate/rule/pds4/FindUnreferencedIdentifiers.java#L53-L119):
1. outer loop running over all targets
2. inner loop running over all targets again but different list
  a. uses if to see if var from (1) is equal to (2)
  b. cannot look up using hash because misuse of overly generic Collection<> instead of Set<> or Map<> or something better suited to task

This part of the job will take awhile and I have no way to estimate it for you. It does not seem to be stuck though.

### 🎉 Integration & Test

_No response_

al-niessner:
## 🗒️ Summary
Multi step process of cleaning up validate to improve performance:
1. Fixed ValidateTarget to be compliant with Java: hashCode() returned different values when equals() returned true.
2. Fixed ValidateTarget.cachedTargets was implemented in a Utility class so very few creators of ValidateTarget used it. Moved cache to ValidateTarget, privatized constructors, used factory methods (build) for consistent use of cache.
3. Fixing Identifier to be compliant with Java: hashCode() returned different values when equals() returned true.
4. Fix all of the loops to use contains() instead. Those loops are a consequence of (3).

## ⚙️ Test Data and/or Report
Automated unit/regression tests below should pass.

## ♻️ Related Issues
Closes #969

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/validate/ValidationTarget.java
+++ b/src/main/java/gov/nasa/pds/tools/validate/ValidationTarget.java
@@ -15,6 +15,7 @@ package gov.nasa.pds.tools.validate;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.HashMap;
 import org.apache.commons.io.FilenameUtils;
 import gov.nasa.pds.tools.util.Utility;
 
@@ -22,6 +23,11 @@ import gov.nasa.pds.tools.util.Utility;
  * Represents a location within a validation subtree that can have errors reported against it.
  */
 public class ValidationTarget implements Comparable<ValidationTarget> {
+  // A static cache of the ValidationTargets.
+  // There is no need to re-evaluate and/or create these
+  // as validation proceeds, as they are static things like
+  // a file or a URL.
+  public static HashMap<String, ValidationTarget> cachedTargets = new HashMap<>();
 
   private TargetType type;
   private String name;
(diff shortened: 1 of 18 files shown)


src/main/java/gov/nasa/pds/tools/validate/ValidationTarget.java
```diff
@@ -189,8 +190,7 @@ public String toString() {
   @Override
   public int hashCode() {
     if (knownHashCode == 0) {
-      String combined = location + ":" + (type == null ? "" : type.toString());
-      knownHashCode = combined.hashCode();
+      knownHashCode = location.hashCode();
```
nutjob4life, line 193: Is there ever a risk that `location.hashCode()` would be `0`, and thus re-compute the hash code?

I guess it doesn't really matter, though … hash codes aren't tremendously expensive

nutjob4life (approved):
LGTM 👍

src/main/java/gov/nasa/pds/tools/validate/ValidationTarget.java
al-niessner: Yes. Not in infinite loop but will recompute 0 every time. I left the knownHashCode for nostalgia. Besides location is a string and I bet Java keeps a static value of it around just like it does the string since they are immutable. Really should just return location.hashCode() and not try to buffer it but I am a nostalgic person.

nutjob4life: Hey, a little nostalgia now and then is just fine 😌

al-niessner:
Unfortunately, it did not improve the unit testing time. Must be dominated by something else.

PR #1077, opened 2024-12-02
Update XXE prevention code

jordanpadams:
## 🗒️ Summary
The feature setting is an older method and is no longer working. Per https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html, it looks like only this line is needed.

```
factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
```

## ⚙️ Test Data and/or Report
Verify tests execute successfully.


jordanpadams:
Merging since tests completed successfully and this is blocking some testing on the pds4-information-model repository. Will create new issue if there are issues that arise from PR review

PR #1078, opened 2024-12-02
Fix code scanning alert no. 442: Resolving XML external entity in user-controlled data

jordanpadams:
Fixes [https://github.com/NASA-PDS/validate/security/code-scanning/442](https://github.com/NASA-PDS/validate/security/code-scanning/442)

To fix the problem, we need to disable external entity resolution in the `TransformerFactory` used in the `SchematronTransformer` class. This can be done by setting specific features on the `TransformerFactory` instance to disallow DTDs and external entities. This will prevent XXE attacks by ensuring that the XML parser does not process external entities.

The changes should be made in the `buildIsoTransformer` method of the `SchematronTransformer` class. We will set the necessary features on the `TransformerFactory` instance to secure it against XXE attacks.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
+++ b/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
@@ -21,6 +21,7 @@ import java.io.StringWriter;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.XMLConstants;
 import javax.xml.transform.Source;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerConfigurationException;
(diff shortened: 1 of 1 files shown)


jordanpadams:
@al-niessner can you review this to make sure it makes sense to you?

al-niessner (approved):
The code makes sense. What does not make sense is why. The user is just processing their own data on their own machine. It is not like there is some weird injection going from a user clicking on a funky link (phishing). In other words, they could blow up their machine by injecting malicious code or they could just take a hammer to it. Hammer seems more direct and likely than some convoluted XXE. Probably never even realize XXE was an option, but this code will force the hammer approach.

jordanpadams:
@al-niessner I agree with the why, but, in the future, if validate is deployed under the hood of some web app, this could potentially eliminate a risk.

PR #1167, opened 2025-03-09
Add more complete JPEG check

### LINKED ISSUE #1118: JPEG validation does not allow JPEG files with trailers
### Checked for duplicates

Yes - I've already checked

### 🐛 Describe the bug

When I validated Product_Ancillary products with Encoded_Images in JPEG format, I noticed that the validator rejected the images as invalid JPEGs.
Investigating the issue revealed that these images have additional data beyond the EOI marker. These images contain a complete JPEG image and will still display correctly. 

The test that the validator used was to check the beginning and the end of each file for the byte sequences 0xFFD8 (SOI (Start-of-interchange)) and 0xFFD9 (EOI (End-of-interchange)). While many JPEG files will pass this test, other JPEG files will not, and this test is inappropriate.

According to the PDS4 Information Model, the Encoding Standard Id of JPEG specifies that "the encoded image is governed by ISO/IEC 10918-1". However, 
ISO/IEC 10918-1 does not specify a file format, only an interchange format. One file format for a JPEG file is described by the JFIF specification (ISO/IEC 10918-5) which is not referenced in the PDS4 definition. (There are also other specifications such as EXIF that are considered to be JPEG files, but this only emphasises that the format is underspecified.)

A compliant JPEG decoder (not necessarily JFIF) will not fail on a JPEG file that contains data after the EOI marker, since it still contains a complete JPEG interchange. ISO/IEC 10918-1 instructs the decoder to stop processing once the EOI is encountered.

### 🕵️ Expected behavior

I expected the validator to emit a WARNING, at most.

The validator should not automatically fail if the EOI is not at the end of the file. It would be better to process the interchange within the file and determine that it is valid. This should have a low impact on performance, since every marker except for SOI and EOI include length indicators, so it's possible to skip most of the file. Since it could still be an indicator of a problem if there is data beyond the EOI, a warning should still be emitted when trailing data is present. Parsing the file could be a net gain for the reliability of the validator, since there is more to the validity of a JPEG beyond the presence of the SOI and EOI.

Additionally, when offset and object length attributes are provided, the validator should use these values, and check for the SOI and EOI markers at the specified locations. Presumably, the offset and object length would define the boundaries of an interchange, so this would be an appropriate test.

I've included a flowchart for reference.

Image

### 📜 To Reproduce

Run validate on the provided label:

`~/bin/validate-3.6.3/bin/validate 20170303t022534s621_sto_l0.b.xml`

Observe the following error:

```
ERROR  [error.file.not_jpeg_compliant]   file:/Users/jessestone/Desktop/orex.tagcams.jpegs/20170303t022534s621_sto_l0.jpg is not valid JPEG file
```

### 🖥 Environment Info

Version of this sofware: 3.6.3
Operating System: MacOS 14.6.1

gov.nasa.pds:validate
Version 3.6.3
Release Date: 2024-11-19 00:03:18

[Archive.zip](https://github.com/user-attachments/files/18623055/Archive.zip)

### 🦄 Related requirements

PDS4 Information Model encoding_standard_id definition: https://pds.nasa.gov/datastandards/documents/im/v1/index_1N00.html#value_pds_encoded_image_pds_encoding_standard_id_jpeg

"SOME LITTLE-KNOWN ASPECTS OF THE HISTORY OF THE JPEG STILL PICTURE-CODING STANDARD, ITU-T T.81 |ISO/IEC 10918-1 (1986-1993)" https://www.itu.int/dms_pub/itu-s/opb/journal/S-JOURNAL-ICTS.V3I1-2020-13-PDF-E.pdf

ISO 10918-1: https://www.w3.org/Graphics/JPEG/itu-t81.pdf

ISO 10918-5: https://www.ijg.org/files/T-REC-T.871-201105-I!!PDF-E.pdf

Known JPEG Trailers: https://exiftool.org/TagNames/JPEG.html

ImageUtil.java: https://github.com/NASA-PDS/validate/blob/5aa2b0475a9c4b4ec6997076f705f40235131f55/src/main/java/gov/nasa/pds/tools/util/ImageUtil.java#L92

### Acceptance Criteria

**Given** 
**When I perform** 
**Then I expect** 

### ⚙️ Engineering Details

_No response_

### 🎉 Integration & Test

_No response_

al-niessner:
## 🗒️ Summary
Use ImageIO if simple JPEG check fails.

## ⚙️ Test Data and/or Report
All unit tests below pass.

## ♻️ Related Issues
Closes #1118 if using ImageIO is the desired way to solve this problem.


jordanpadams (approved):

jordanpadams (approved):
Looks good.

PR #1266, opened 2025-06-04
detect inventory underflow when using pds4.label rule

### LINKED ISSUE #1234: validate does not raise a warning when a table has more records than label says it has
### Checked for duplicates

Yes - I've already checked, e.g. #149 

### 🐛 Describe the bug

When I ran Validate with the `--rule pds4.bundle` option, it reported a collection file with a mismatch between the number of records reported in the label compared to the actual number of records in the table:

```
ERROR  [error.table.records_mismatch]   Number of records read is not equal to the defined number of records in the collection (expected 5, got 7).
```

(and that was very helpful). However, when I ran Validate at the individual product level (no `--rule` setting) on the same collection inventory file, I noticed that Validate passed the label without any warning/error even though the label had only 5 `<records>` for a table that actually contained 7 records. I confirmed that the same behaviour occurred with a non-collection product (I attach the `collection_document.xml`, and another example product label, `vg2u_49xr_1986025t024210.xml` with their associated files below).

Changing the value of `<records>` in the label and re-running Validate on that collection file (with no `--rule` setting), I saw that `error.table.records_mismatch` is only raised when the label `<records>` is greater than what's actually in the table:
```
 ERROR  [error.table.records_mismatch]   data object 1: Number of records read is not equal to the defined number of records in the label (expected 8, got 7).
```

This is related to other issues, e.g. #149, [where it seems that the desired outcome](https://github.com/NASA-PDS/validate/issues/149#issuecomment-616769787) was for:
– an error to be raised if the table is shorter than the label says it is (e.g. table has 7 records but label says 8), 
– a warning to be raised if the table is longer than the label says it is  (e.g. table has 7 records but label says 5).

However, given that was back in 2020, I'd like to also check that that is still the desired behaviour?

Tagging interested parties, @radiosci and @matthewtiscareno. 

### 🕵️ Expected behavior

I expected Validate to report a warning (or perhaps an error) if the number of `<records>` in the label is fewer than the actual number of records in the table.
(Validate does report an error if the number of `<records>` described in the label is more than what's actually in the table.)

### 📜 To Reproduce

```
% validate ./collection_document.xml
```
```
% validate ./vg2u_49xr_1986025t024210.xml
```

### 🖥 Environment Info

- Validate v3.7.0 (v3.6.3 and v3.5.1 also exhibit this behaviour)
- MacOS Sequoia 15.1

3.7.0, 3.6.3, 3.5.1

tar.gz too large to attach directly, [so have put on Dropbox](https://www.dropbox.com/scl/fi/94msca0uot12seorrzpjn/vg2u_data.tar.gz?rlkey=dledz455o2a8r9agan4ce3rb0&st=6d11eomx&dl=0)

### 🦄 Related requirements

🦄 #xyz

### ⚙️ Engineering Details

_No response_

### 🎉 Integration & Test

_No response_

al-niessner:
## 🗒️ Summary
Targeted change to add a check to inventory only tables for over/under flow when using pds4.label rule.

## ⚙️ Test Data and/or Report
All automated tests below should pass. Existing tests 217 and 173 check for underflow already and should pass.

## ♻️ Related Issues
Closes #1234


al-niessner:
@jordanpadams 

The existing test for #173 and #217 already check inventory tables for underflow. Updated the expectations to include it and they all pass now.

nutjob4life (commented):
I got an error during `mvn test`:
```
org.opentest4j.AssertionFailedError: summary:productValidation:failed ==> expected: <0> but was: <1>
	at cucumber.StepDefs.compare_to_the(StepDefs.java:182)
	at ✽.compare to the expected outcome "summary:totalErrors=2,summary:totalWarnings=5,summary:referentialIntegrity:failed=1,summary:messageTypes:error.table.records_mismatch=2,summary:messageTypes:warning.integrity.reference_not_found=5".(file:///Users/kelly/Documents/Clients/JPL/PDS/Development/nasa-pds/validate/src/test/resources/features/pre.3.6.x.feature:5)
```
Is this a known or expected one?

al-niessner:
Tested before oops push?

nutjob4life (approved):
Haha I guess so @al-niessner! Here's what I get now:
```
[INFO] Tests run: 261, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 630.2 s -- in cucumber.CucumberTest
[INFO] 
[INFO] Results:
[INFO] 
[WARNING] Tests run: 284, Failures: 0, Errors: 0, Skipped: 23
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

jordanpadams:
@al-niessner this is fine, but this only partially solves https://github.com/NASA-PDS/validate/issues/1234.

if we can, let's try to [implement this?](https://github.com/NASA-PDS/validate/issues/1234#issuecomment-2932472680)

jordanpadams (approved):

Issue #1317, opened 2025-07-24
Validate memory shortage


ralanis-jpl:
### Checked for duplicates

No - I haven't checked

### 🐛 Describe the bug

Validate 3.7.0 ran out of memory while running an integrity check on a bundle.

### 🕵️ Expected behavior

I expected Validate to not run out of memory.

### 📜 To Reproduce

1. Here is the command I used, 

$ ./validate /export/mslwork/MSLNAV_1XXX -R pds4.bundle --skip-product-validation -v 2 --disable-context-mismatch-warnings -r /home/alanis/bundle_39_draft_MSLNAV_1XXX_rpt_01.rpt

2. This is the error that came up on the screen,

Caused by: java.lang.OutOfMemoryError: Java heap space
Completed execution in 78140316 ms

3. grep results,

$ grep ERROR bundle_39_draft_MSLNAV_1XXX_rpt_01.rpt | sort | uniq
   ERROR [error.validation.internal_error] Error executing validationTask: java.lang.OutOfMemoryError: Java heap space
   ERROR [error.validation.internal_error] Uncaught exception while validating: null

A portion of the output report:

  FAIL: file:/export/mslwork/MSLNAV_1XXX/EXTRAS/
      ERROR  [error.validation.internal_error]   Uncaught exception while validating: null

  FAIL: file:/export/mslwork/MSLNAV_1XXX/
      ERROR  [error.validation.internal_error]   Error executing validationTask: java.lang.OutOfMemoryError: Java heap space

Summary:

  835002 product(s)
  46 error(s)
  0 warning(s)

  Product Validation Summary:
    0          product(s) passed
    0          product(s) failed
    835002     product(s) skipped
    835002     product(s) total

  Referential Integrity Check Summary:
    1          check(s) passed
    0          check(s) failed
    0          check(s) skipped
    1          check(s) total

  Message Types:
    46           error.validation.internal_error

End of Report
...

### 🖥 Environment Info

- Version of this software [3.7.0]
- Operating System: [Unix, mslpds1:]
...

### 📚 Version of Software Used

3.7.0

### 🦄 Related requirements

🦄 #xyz

jordanpadams:
@ralanis-jpl can you try increasing the heap size and see if that fixes it? https://nasa-pds.github.io/validate/operate/errors.html#outofmemory

jordanpadams:
Closing this as stale

Issue #931, opened 2024-06-11
Referential integrity check takes much longer than it seems it should


rgdeen:
### Checked for duplicates

No - I haven't checked

### 🐛 Describe the bug

"bug" Is a strong word, but it's the closest category.

I have a bundle (MSAM2) with ~2 million products in it.  I can do product-level validation in parallel using KDP or Nucleus or other technology to farm it out to a bunch of nodes.  However, referential integrity (verifying the inventory files are correct and match the files present) has to be done on the bundle as a whole - I'm not aware of any way to split that up.  (maybe by collection, but there are only 2 relevant collections here so that doesn't help much).

In order to do this, I'm running with product and content validation turned off.  But it is still taking an inordinate amount of time.  As of this writing, it's been running 6 days and per the log has gotten through 1,047,476 out of 2,013,873 products - about halfway.  That's a rate of about 2 per second.  Seems like it should be able to do better in this case.

### 🕵️ Expected behavior

Well I expected what I got ;-) but I would hope the RI checks could be faster.

### 📜 To Reproduce

Here's the command line:

```
/path/to/msam2/validate-3.5.1/bin/validate -target /path/to/msam2/annex_ehlmann_caltech_msl_msam2 --report-file bundle.valrpt -R pds4.bundle --skip-content-validation --skip-product-validation
```

### 🖥 Environment Info
```
$ uname -a
Linux machine-name 3.10.0-1160.76.1.el7.x86_64 #1 SMP Tue Jul 26 14:15:37 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

$ java -version
java version "17.0.11" 2024-04-16 LTS
Java(TM) SE Runtime Environment (build 17.0.11+7-LTS-207)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.11+7-LTS-207, mixed mode, sharing)
```

### 📚 Version of Software Used

```
$ /mnt/pdsdata/scratch/rgd/msam2/validate-3.5.1/bin/validate -version

gov.nasa.pds:validate
Version 3.5.1
Release Date: 2024-05-25 17:45:47

Copyright 2019, by the California Institute of Technology ("Caltech").
All rights reserved.
```

al-niessner:
@rgdeen Can you tell me if it is slowing over time? We previously fixed slowness for #824, which should be part of the validate you are using, and knowing if it is slowing may help shorten the fix time. Could be a red herring too.

al-niessner:
[...]

PR #1326, opened 2025-08-05
Various fixes that should improve validate peformance

### LINKED ISSUE #488: Referential integrity check does not complete with large bundles.
## 🐛 Describe the bug
When validating a large bundle, there is a risk that the referential integrity check will not complete, due to an out of memory error. This happened recently with the spacewatch bundle, which has 1.8 million products.

## 📜 To Reproduce

Run validate on the spacewatch bundle:
`validate -D -R pds4.bundle -t gbo.ast.survey.spacewatch`

[spacewatch_labels.txt.gz](https://github.com/NASA-PDS/validate/files/8516820/spacewatch_labels.txt.gz)

I've attached the output from our last validation run which illustrates the error, as well.

## 🕵️ Expected behavior
The referential integrity check should have completed, and raised errors in the bundle related to a LID collection id mismatch. At minimum, a more clearly marked abnormal exit would be desirable, so that it's clear that the validation results cannot be used.

## 📚 Version of Software Used
Validate 2.1.4

## 🩺 Test Data / Additional context

If needed, I can provide a stub bundle with all of the spacewatch labels.

##  🏞Screenshots
Screen Shot 2022-04-19 at 3 41 23 PM

## 🖥 System Info
 - OS: CentOS7

---

jstone-psi:
### Overview

The current version of validate may perform acceptably for smaller jobs. However, for tasks such as referential integrity validation, the validator cannot currently be used successfully for very large jobs such as surveys. The memory and computation requirements make the job infeasible. Some alternatives to running the validator that have been proposed include running a registry-based validation, and using other toolchains to verify referential integrity. In addition to these techniques, I looked at the code of the validator, and made a series of improvements outlined below. The changes to the code caused the validation task to run over an order of magnitude faster. A job that did not complete after six months now completes over a weekend. While the memory consumption is still high, it now completes with a lower peak memory usage than before.

The changes made pass all unit tests that were in validate-3.8.0-SNAPSHOT at the time of the changes.

#### Instrumentation Improvements

* Now logs any validation rule that takes over a second when in debug mode.

#### Bundle referential integrity checks

* Brute force referential integrity check now checks against a set of LIDs, so that the check no longer takes linear time.
* Changed additional array lists into sets where appropriate, so that membership tests no longer take linear time.

#### Collection referential integrity checks

In-Memory target registrar now maintains an index of LIDs to declarations and references, so that you can check only references or declarations with the same LID for near-neighborness

#### General perforamnce improvements

* No longer creates a new JAXB context for every label. This is an upstream issues in pds4-jparser, which I'm not equipped to rebuild. I tempoararily pulled in the classes from jparser to make these improvements. It would make sense to push these back upstream.
* XMLExtractor now only parses the document once, instead of every time a value is retrieved
* Saved the result of isDir, so that it doesn't get called several times in a row.

#### Memory consumption
* Converted some hashmap caches to a guava cache, using soft references. This will free up additional memoery as needed.
* Now builds up set differences using a stream, rather with than removeAll, since that would build up a set that was the same size as the original set.
* Removed some redundant data structures from the In-Memory target registrary

#### General performance recommendations

* The validate job will run even faster when the labels are on an SSD. Referential integrity could be performed with only the labels. For a full validation, the labels could be layered using OverlayFS.
* The memory consumption of validate seems to be 100% of the total label volume. Having less memory will likely work with the changes made above, but it may cause labels to be reparsed. This is less of an issue for SSD-based storage.

#### Other notes

* Ut's also possible to do a quick check with bash scripts to verify that the labels are likely to pass referential integrity validation. I am working on the scripts and a process to do this.
* Further performance gains could likely be achieved by updating the code so that it can be parallelized. In fact, I saw some attempts at parallelization when I was reviewing the code. These did not work when I activated them, likely due to side-effects.

## ⚙️ Test Data and/or Report
One of the following should be included here:
Here's a sample run of NEAT quest, which previously did not complete after running for months.
The report was too large to upload, linked here:

https://oort.psi.edu/index.php/s/Abi6y4mz462SnXG

## ♻️ Related Issues

Resolves https://github.com/NASA-PDS/validate/issues/488.

This should also help with:
[Validate memory shortage](https://github.com/NASA-PDS/validate/issues/1317)
[Referential integrity check takes much longer than it seems it should](https://github.com/NASA-PDS/validate/issues/931)


jordanpadams:
thanks @jstone-psi . I will need to create a new branch to merge these changes into so the tests can build.

PR #1386, opened 2025-10-07
Upgrade to Java 17 to support required dependency upgrades

al-niessner:
## 🗒️ Summary
Requires JUnit 5 and Java 17. Requires eclipse update to build path for JUnit 5 and possibly Java 17.

## ⚙️ Test Data and/or Report
All unit tests below pass.

## ♻️ Related Issues
NA

### DIFF
--- a/.github/workflows/branch-cicd.yaml
+++ b/.github/workflows/branch-cicd.yaml
@@ -33,7 +33,7 @@ jobs:
 
         strategy:
             matrix:
-                java-version: [11, 17]
+                java-version: [17]
 
         steps:
             -
--- a/.github/workflows/stable-cicd.yaml
+++ b/.github/workflows/stable-cicd.yaml
@@ -72,7 +72,7 @@ jobs:
                 uses: NASA-PDS/roundup-action@stable
                 with:
                     assembly: stable
-                    packages: openjdk11-jdk
+                    packages: openjdk17-jdk
                     maven-build-phases: install
                     maven-doc-phases: clean,site,site:stage
                     maven-stable-artifact-phases: clean,site,site:stage,deploy
--- a/.github/workflows/unstable-cicd.yaml
+++ b/.github/workflows/unstable-cicd.yaml
@@ -75,7 +75,7 @@ jobs:
                 uses: NASA-PDS/roundup-action@stable
                 with:
                     assembly: unstable
-                    packages: openjdk11-jdk
+                    packages: openjdk17-jdk
                     maven-build-phases: install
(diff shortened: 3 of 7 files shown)


al-niessner:
@jordanpadams 

Here is your rollback branch. When you are ready, just merge it. I recommend sooner rather than latter because the latest cucumber is breaking without it as JUnit 4 support is getting flaky.

nutjob4life (approved):
LGTM: 👍 

Also:
```
[WARNING] Tests run: 23, Failures: 0, Errors: 0, Skipped: 23
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  13:24 min
[INFO] Finished at: 2025-10-07T10:40:03-05:00
```

jordanpadams:
@al-niessner I do not feel comfortable upgrading to Java 17 only at this point, until we can confirm all of our Java applications are upgraded to Java 17. Can you confirm there is not other workaround to get cucumber to work?

al-niessner:
@jordanpadams 

That is fine.

registry-common: 11 (backward looking usually works)
regsitry-mgr: 17
harvest: 17 and 21

There may be a way to keep the old crufty stuff, but cucumber-vintage-engine is no longer keeping pace. I am sure we can expend effort to fall further behind to spend more effort to catch up in the future if that is desire.

When you are comfortable with it, then merge this PR. Delete this PR if you never want to update.

jordanpadams:
@al-niessner copy. If we make this upgrade, can we make sure we update the documentation as well?

al-niessner:
@jordanpadams 

I did a scan of all files in validate looking for anything like java 11 or jdk 11 or jre 11 and only found the stuff in .github. I then rooted around in docs a little. I really do not understand it well because I could not find the reference to what Java was required - in other words, thought of it already and could not find anything to update.

jordanpadams:
@al-niessner  https://github.com/NASA-PDS/validate/blob/main/src/site/xdoc/install/index.xml.vm#L87

we also need to make sure pds4-jparser is upgraded

al-niessner:
@jordanpadams 

No you do not. JDKs look backward well. A JDK 17 is happy with JDK 11 classes because it understands them. JDK 11 is not too happy with JDK 17 because it came in the future. It is why this PR uses the current jparser and all tests run successfully. Our tests use and exercise jparser heavily.

jordanpadams:
@al-niessner once we upgrade the docs, I will merge.

PR #1422, opened 2025-11-03
Upgrade apache commons configuration to 2.x

jordanpadams:
## 🗒️ Summary
Upgrade apache commons configuration to 2.x

## ⚙️ Test Data and/or Report
See github actions

## ♻️ Related Issues

- Resolves https://github.com/NASA-PDS/validate/security/dependabot/14

### DIFF
--- a/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
+++ b/src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
@@ -74,7 +74,9 @@ import org.apache.commons.cli.ParseException;
 import org.apache.commons.cli.UnrecognizedOptionException;
 import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.PropertiesConfiguration;
-import org.apache.commons.configuration2.builder.fluent.Configurations;
+import org.apache.commons.configuration2.builder.FileBasedConfigurationBuilder;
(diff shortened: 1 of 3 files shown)


al-niessner (dismissed):

al-niessner (changes requested):
I can do it if you do not want to, but two of the test use the config option and they are failing because the report writer is never being generated. I am guessing that the config is not being processed quite right.

src/main/java/gov/nasa/pds/validate/ValidateLauncher.java
```diff
@@ -706,8 +706,8 @@ private void copyFile(File source, File dest) throws IOException {
    */
   public void query(File configuration) throws ConfigurationException {
     try {
-      AbstractConfiguration.setDefaultListDelimiter(',');
-      Configuration config = new PropertiesConfiguration(configuration);
+      Configurations configs = new Configurations();
```
al-niessner: Did the , become the default or did we forget to set it with the config2?

jordanpadams:
closing this PR. we are using trusted configurations so this is a very low risk

jordanpadams:
@al-niessner decided to not move forward with this upgrade since it was breaking something, and not sure what, and this is a very low risk

Issue #1567, opened 2026-04-01
As a PDS developer, I want static mutable utility singletons converted to instance-based classes so that parallel label validation becomes possible


jordanpadams:
## Checked for duplicates
Yes - I've already checked

## 🧑‍🔬 User Persona(s)
Data Engineer, PDS Developer

## 💪 Motivation
...so that I can enable true in-JVM parallel label validation by eliminating the global lock contention caused by static mutable state across utility classes.

## 📖 Additional Details

**Impact:** High (prerequisite for parallelism work)

**Relevant files:**
- `src/main/java/gov/nasa/pds/tools/util/LabelUtil.java` (lines 42-82, 267-268)
- `src/main/java/gov/nasa/pds/tools/util/ReferentialIntegrityUtil.java` (lines 48-132)
- `src/main/java/gov/nasa/pds/tools/validate/CrossLabelFileAreaReferenceChecker.java` (lines 16-24)
- `src/main/java/gov/nasa/pds/tools/util/EveryNCounter.java` (lines 1-27)

**Problem:**
Multiple utility classes use `static` mutable fields as global state, requiring `synchronized` access and preventing concurrent processing:

- **`LabelUtil`**: All methods are `static synchronized`, with static fields for `informationModelVersions`, `location`, `report`, etc.
- **`ReferentialIntegrityUtil`**: Entirely static state — `lidvidsCumulative`, `lidsCumulative`, `lidOrLidVidReferencesCumulative`, multiple `HashMap`s, etc.
- **`CrossLabelFileAreaReferenceChecker`**: Static `HashMap` fields with no synchronization.
- **`EveryNCounter`**: Singleton with `synchronized` on enum constants.

**Recommendation:**
Convert these from static singletons to instance-based classes passed through a validation context object. Also improves testability, since static state currently requires explicit `reset()` calls between test runs.

## Acceptance Criteria
**Given** a multi-threaded validation run
**When I perform** concurrent label validation
**Then I expect** no global lock contention from static utility state

Issue #1566, opened 2026-04-01
As a PDS data engineer, I want label validation to run in parallel within a single JVM so that I can fully utilize available CPU cores without spawning separate processes


jordanpadams:
## Checked for duplicates
Yes - I've already checked

## 🧑‍🔬 User Persona(s)
Data Engineer, Node Operator

## 💪 Motivation
...so that I can validate large bundles in a fraction of the current time by utilizing multiple CPU cores within one JVM, eliminating the need for the heavy-handed `validate-bundle` shell script that spawns separate JVM processes via GNU Parallel.

## 📖 Additional Details

**Impact:** Very High

**Relevant files:**
- `src/main/java/gov/nasa/pds/tools/label/LabelValidator.java` (line 456)

**Problem:**
`LabelValidator.parseAndValidate()` is declared `synchronized`, meaning only one thread can validate a label at a time within a single JVM. This completely negates any in-process parallelism. The `synchronized` keyword exists because the method mutates shared instance state (`cachedParser`, `cachedValidatorHandler`, `cachedSchematron`, etc.).

**Recommendation:**
Refactor `LabelValidator` to be thread-safe by making the mutable state (parser, validator handler, schematron cache) either thread-local or passed as method parameters.

Note: depends on resolving the pervasive static mutable state first (see companion issue on `LabelUtil`, `ReferentialIntegrityUtil`, etc.).

## Acceptance Criteria
**Given** a bundle with many product labels
**When I perform** validation on a multi-core machine
**Then I expect** labels to be validated concurrently within a single JVM process

PR #1572, opened 2026-04-01
Improve validation performance by caching compiled schematron Templates to avoid repeated XSLT compilation

### LINKED ISSUE #1565: As a PDS data engineer, I want schematron transformers to be cached rather than recompiled for every label so that bundle validation is significantly faster
## Checked for duplicates
Yes - I've already checked

## 🧑‍🔬 User Persona(s)
Data Engineer, Node Operator

## 💪 Motivation
...so that I can validate large bundles with many products without paying the cost of XSLT recompilation on every single label, reducing validation time by an estimated 30–50%.

## 📖 Additional Details

**Impact:** Very High

**Relevant files:**
- `src/main/java/gov/nasa/pds/tools/label/LabelValidator.java` (lines 611-622)
- `src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java` (lines 112-117)

**Problem:**
`SchematronTransformer.transform(String)` is called for every label inside the schematron validation loop in `LabelValidator.parseAndValidate()`. Each call re-parses the schematron string into a `Transformer` via XSLT compilation — an extremely expensive operation. The schematron string content is cached (`cachedSchematron`), but the compiled `Transformer` object is not. A new `StreamSource(new StringReader(source))` is created and the full XSLT compilation pipeline runs every time.

**Recommendation:**
Cache the compiled `Transformer` (or `Templates`) object keyed by the schematron string hash, rather than recompiling from the string on every invocation.

## Acceptance Criteria
**Given** a bundle with many product labels sharing the same schematron
**When I perform** validation
**Then I expect** the schematron XSLT is compiled once and reused, not recompiled per label

jordanpadams:
## 🗒️ Summary

Adds a `HashMap<String, Templates>` cache to `SchematronTransformer` so the expensive ISO schematron XSLT compilation runs once per unique schematron source string rather than once per label. Subsequent calls to `transform(String)` return a new `Transformer` from the pre-compiled `Templates` object — eliminating the dominant CPU cost during bundle validation for large collections with many products.

Changes in `SchematronTransformer.java`:
- Added `cachedTemplates` instance field (`Map<String, Templates>`)
- Extracted XSLT compilation into private `compileSchematron()` returning `Templates`
- Cache lookup/store added to `transform(String, ProblemHandler)`
- Added `clearCache()` for explicit lifecycle management
- Added debug logging for cache hits/misses

### 🤖 AI Assistance Disclosure
- [ ] No AI assistance used
- [ ] AI used for light assistance (e.g., suggestions, refactoring, documentation help, minor edits)
- [ ] AI used for moderate content generation (AI generated some code or logic, but the developer authored or heavily revised the majority)
- [x] AI generated substantial portions of this code

Estimated % of code influenced by AI: 90%

## ⚙️ Test Data and/or Report

Existing Cucumber integration tests in `src/test/resources/features/` cover schematron-based validation. No new test data required — this change is transparent to callers. Performance improvement observable via timing on any bundle validation run.

## ♻️ Related Issues

Fixes #1565

## 🤓 Reviewer Checklist

*Reviewers: Please verify the following before approving this pull request.*

### Documentation and PR Content
- [ ] **Documentation:** README, Wiki, or inline documentation (Sphinx, Javadoc, Docstrings) have been updated to reflect these changes.
- [x] **Issue Traceability:** The PR is linked to a valid GitHub Issue
- [x] **PR Title:** The PR title is "user-friendly" clearly identifying what is being fixed or the new feature being added, that if you saw it in the Release Notes for a tool, you would be able to get the gist of what was done.

### Security & Quality
- [ ] **SonarCloud:** Confirmed no new High or Critical security findings.
- [ ] **Secrets Detection:** Verified that the Secrets Detection scan passed and no sensitive information (keys, tokens, PII) is exposed.
- [ ] **Code Quality:** Code follows organization style guidelines and best practices for the specific language (e.g., PEP 8, Google Java Style).

### Testing & Validation
- [ ] **Test Accuracy:** Verified that test data is accurate, representative of real-world PDS4 scenarios, and sufficient for the logic being tested.
- [ ] **Coverage:** Automated tests cover new logic and edge cases.
- [ ] **Local Verification:** (If applicable) Successfully built and ran the changes in a local or staging environment.

### Maintenance
- [ ] **Backward Compatibility:** Confirmed that these changes do not break existing downstream dependencies or API contracts (or that breaking changes are clearly documented).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [«unknown»](https://github.com/NASA-PDS/validate/tree/«unknown») (2026-03-30)
+## [«unknown»](https://github.com/NASA-PDS/validate/tree/«unknown») (2026-04-01)
 
 [Full Changelog](https://github.com/NASA-PDS/validate/compare/v4.0.8...«unknown»)
 
@@ -12,6 +12,8 @@
 
 **Defects:**
 
+- Validate does not count cleanly-passing products, incorrectly triggering no\_products\_found error [\#1557](https://github.com/NASA-PDS/validate/issues/1557) [[s.medium](https://github.com/NASA-PDS/validate/labels/s.medium)]
+- validate does not fail or report errors when an explicitly-specified target file does not exist [\#1548](https://github.com/NASA-PDS/validate/issues/1548) [[s.medium](https://github.com/NASA-PDS/validate/labels/s.medium)]
 - error.directory.unallowed\_name not raised on Windows for absolute directory\_path\_name [\#1539](https://github.com/NASA-PDS/validate/issues/1539) [[s.medium](https://github.com/NASA-PDS/validate/labels/s.medium)]
 - When validate reports a directory has a warning it says "XML could not be parsed." which is misleading. [\#1510](https://github.com/NASA-PDS/validate/issues/1510) [[s.medium](https://github.com/NASA-PDS/validate/labels/s.medium)]
 - Referential integrity not being checked in all places there is an Internal\_Reference area [\#1481](https://github.com/NASA-PDS/validate/issues/1481) [[s.medium](https://github.com/NASA-PDS/validate/labels/s.medium)]
--- a/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
+++ b/src/main/java/gov/nasa/pds/tools/label/SchematronTransformer.java
@@ -20,8 +20,11 @@ import java.io.StringReader;
 import java.io.StringWriter;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.XMLConstants;
 import javax.xml.transform.Source;
@@ -47,7 +50,7 @@ import gov.nasa.pds.tools.validate.ProblemHandler;
  */
 public class SchematronTransformer {
   private static final Logger LOG = LoggerFactory.getLogger(SchematronTransformer.class);
-  private final Map<String, Templates> cachedTemplates = new HashMap<>();
+  private final Map<String, Templates> cachedTemplates = new ConcurrentHashMap<>();
   
   /**
    * Constructor.
@@ -101,6 +104,9 @@ public class SchematronTransformer {
    *
    * @throws TransformerException If an error occurred during the transform process.
    */
+  // Not cached — Source objects are not stably comparable/hashable.
+  // In practice LabelValidator only calls the String-based overload (line 622),
+  // so this uncached path does not affect bundle-validation performance.
   public Transformer transform(Source source, ProblemHandler handler) throws TransformerException {
     return compileSchematron(source, handler).newTransformer();
   }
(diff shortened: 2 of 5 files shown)


jordanpadams (commented):
## PR Review

**Overall: Approve with minor comments.** The implementation is correct and well-targeted. Using `javax.xml.transform.Templates` is exactly the right JAXP API for this — it represents a compiled, thread-safe stylesheet that produces new `Transformer` instances cheaply. The core fix is sound.

---

### What's good

- **Correct API choice.** `Templates` is JAXP's purpose-built mechanism for compiled stylesheet reuse. `newTransformer()` is lightweight; the expensive compilation (`newTemplates()`) now happens once.
- **Clean extraction.** `compileSchematron()` as a private method reads well and avoids duplication between the cached and uncached code paths.
- **`Transformer` not cached.** Correctly calls `templates.newTransformer()` on each use — `Transformer` is not thread-safe and must not be shared.
- **Debug logging.** Cache hit/miss logging will make performance profiling straightforward.

---

### Comments

**1. `HashMap` is not thread-safe — use `ConcurrentHashMap` as a forward-looking safety measure**

`SchematronTransformer` is currently safe because `LabelValidator.parseAndValidate()` is `synchronized`, so the cache is only accessed by one thread at a time. But the parallel-validation work (#1566, #1567) will remove that `synchronized` constraint. Switching to `ConcurrentHashMap` now costs nothing and avoids a subtle data race regression later:

```java
// current
private final Map<String, Templates> cachedTemplates = new HashMap<>();

// safer
private final Map<String, Templates> cachedTemplates = new ConcurrentHashMap<>();
```

**2. Cache key is the full schematron string — memory overhead for large schematrons**

Using the source string itself as the key means the map holds a reference to the entire schematron text in addition to the compiled `Templates`. For large schematrons this doubles memory use for cached entries. A minor improvement would be keying on a hash of the source (e.g., SHA-256 hex) rather than the full string. Low priority but worth knowing.

**3. `clearCache()` is public but not wired up anywhere — consider package-private or remove**

The commit message says the cache "naturally resets when `LabelValidator.clear()` creates a new instance." If that's the lifecycle, `clearCache()` appears to be dead code. Either wire it to `LabelValidator.clear()` or make it package-private to limit surface area. If it's intended for future use, a `// for testing` comment would clarify intent.

**4. `transform(Source, ProblemHandler)` bypass — cache is not used for all call paths**

The `Source`-based `transform(Source, ProblemHandler)` path bypasses the cache entirely — it calls `compileSchematron()` directly every time. If any caller goes through the `Source` overload rather than the `String` overload, they get no benefit. Worth confirming that `LabelValidator` call sites only go through the `String` path in practice.

**5. No test for cache behavior**

There's no test asserting that `compileSchematron` is called only once for repeated identical inputs. A simple unit test on `SchematronTransformer` using a spy or call counter would protect this from regression.

---

### Summary

| Comment | Severity |
|---|---|
| Use `ConcurrentHashMap` for future thread-safety | Recommended |
| Cache key memory overhead | Low / informational |
| `clearCache()` wiring / visibility | Minor |
| `Source`-path cache bypass | Minor — verify not hit in practice |
| No cache-hit regression test | Recommended |

al-niessner:
@jordanpadams 

Becareful with this one. I found that the schematron libraries that we use have hidden globals that cause things to be remembered. Of course that was 20 versions ago...

jordanpadams:
> @jordanpadams
> 
> Becareful with this one. I found that the schematron libraries that we use have hidden globals that cause things to be remembered. Of course that was 20 versions ago...

@al-niessner copy that. I am going to squash these updates to ensure we can easily revert if we determine it is introducing bugs we can't fix.

Issue #1352, opened 2025-09-09
Add the ability to output JSON report validate docker as a single line JSON


ramesh-maddegoda:
### Are you sure this is not a new requirement or bug?

Yes

### Task Type

Sub-task

### 💡 Description

This task should add the ability to output JSON report validate docker as a single line JSON. This will help to have the Validate JOSN report as a single CloudWatch log message, when validate docker image is executed through ECS/ Nucleus.  Without this option, the JSON report is devided in to multiple CloudWatch log messages due to newline characters.

tloubrieu-jpl:
Ramesh updated the validate docker image and testing it in airflow.

tloubrieu-jpl:
Initially the docker image add the json report switch hardcoded. Ramesh made the docker image more flexible.

jordanpadams:
Status: Updated docker compose to get single-line logs to insert into OpenSearch

al-niessner:
@jordanpadams @ramesh-maddegoda 

You should do this by adding a new logging type like JSON-notpretty that does it as one line. It should be trivial as in a half days work at most. The intent of all current logging is to be human readable friendly so altering what is there will have surprise consequences like me complaining that I cannot read the JSON file from a cucumber run.

jordanpadams:
@ramesh-maddegoda @al-niessner I created a new parent theme for this since I think we should take a step back and come up with a consistent logging format / library for supporting Nucleus tracking

https://github.com/NASA-PDS/nucleus/issues/178

ramesh-maddegoda:
Moved this to in-progress, since this is related to https://github.com/NASA-PDS/nucleus/issues/179 and not blocked anymore.

jordanpadams:
@ramesh-maddegoda has a PR been created for this?

ramesh-maddegoda:
> [@ramesh-maddegoda](https://github.com/ramesh-maddegoda) has a PR been created for this?

@jordanpadams , I created PR https://github.com/NASA-PDS/validate/pull/1636

PR #1636, opened 2026-07-02
ADD streaming-json report style for line-delimited product validation

ramesh-maddegoda:
# Add streaming-json report style for line-delimited product validation

Related: https://github.com/NASA-PDS/validate/issues/1352

## 🗒️ Summary

This PR adds a new `streaming-json` report style to the Validate Tool, designed explicitly for high-throughput stream ingestion (one streaming JSON object result per product).

### Key Changes
* **New Report Implementation:** Created `StreamingJsonReport.java` to output individual product validation results as single-line, line-delimited JSON objects rather than wrapping the entire run in a massive JSON array structure.
* **Granular Log Stream Metrics:** Integrated real-time counting logic directly inside the product output blocks. Every product JSON record natively tracks `fatal`, `error`, `warning`, and `info` count totals per line to simplify filtering and dashboards.
* **CLI Flag Routing:** Hooked up `ValidateLauncher.java` to handle `-s streaming-json` and route stream creation/formatting dynamically based on user choices.

### 🤖 AI Assistance Disclosure
- [ ] No AI assistance used
- [ ] AI used for light assistance (e.g., suggestions, refactoring, documentation help, minor edits)
- [] AI used for moderate content generation (AI generated some code or logic, but the developer authored or heavily revised the majority)
- [x ] AI generated substantial portions of this code

Estimated % of code influenced by AI: 90%

---

### Manual Execution & CloudWatch Log Stream Verification
The updated binary was built into a local Docker image, pushed to an Amazon ECR registry, and executed inside an AWS ECS Task by a Nucleus Airflow workflow. 

The tool successfully ran a product batch. The output stream captured directly within the AWS CloudWatch log viewer displays expected formatting per individual product line, concluding with a structural summary record:

```json
{
    "label": "file:/mnt/data/sample-staging-bucket/sample_dir/product_data_file.xml",
    "lidvid": "urn:nasa:pds:sample_bundle:sample_collection:product_data_id::1.0",
    "status": "PASS",
    "fatal": 0,
    "error": 0,
    "warning": 3,
    "info": 0,
    "messages": [
        {
            "severity": "WARNING",
            "type": "warning.label.context_ref_mismatch",
            "message": "Context reference name mismatch. LIDVID: 'urn:nasa:pds:context:instrument:sample_node.sample_instrument'. Value: 'Sample Instrument Name' Expected one of: '[Expected Instrument Name]'",
            "line": 61
        },
        {
            "severity": "WARNING",
            "type": "warning.label.context_ref_mismatch",
            "message": "Context reference name mismatch. LIDVID: 'urn:nasa:pds:context:telescope:sample_node.sample_telescope'. Value: 'Sample Telescope Name' Expected one of: '[Expected Telescope Name 1, Expected Telescope Name 2]'",
            "line": 69
        },
        {
            "severity": "WARNING",
            "type": "warning.label.context_ref_mismatch",
            "message": "Context reference name mismatch. LIDVID: 'urn:nasa:pds:context:facility:observatory.sample_facility'. Value: 'Sample Observatory Facility Name' Expected one of: '[Expected Facility Name]'",
            "line": 77
        }
    ]
}
```


src/main/java/gov/nasa/pds/validate/report/StreamingJsonReport.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    protected void summarizeTotals(int errors, int total, int warnings) {
+        // Intentionally empty: Replaced by target summary record type handled in summarizeProds
+    }
+
+    @Override
+    protected void summarizeProds(int failed, int passed, int skipped, int total) {
+        JsonObject summary = new JsonObject();
+        summary.addProperty("recordType", "summary");
```
nutjob4life, line 124: I don't think I'm understanding: here, we add a `recordType` of `summary` to summaries but we don't do something similar to products, like

    "recordType": "product"

or something?

src/main/java/gov/nasa/pds/validate/report/StreamingJsonReport.java
```diff
@@ -0,0 +1,136 @@
...
+        // Intentionally empty: Context references are skipped to isolate file validation metrics
+    }
+
+    @Override
+    protected void summarizeTotals(int errors, int total, int warnings) {
+        // Intentionally empty: Replaced by target summary record type handled in summarizeProds
+    }
+
+    @Override
+    protected void summarizeProds(int failed, int passed, int skipped, int total) {
```
nutjob4life, line 122: This method receives `int skipped` as a parameter but doesn't use it. Should it go into the summary?
```java
summary.addProperty("skipped", skipped);
```

src/main/java/gov/nasa/pds/validate/report/StreamingJsonReport.java
```diff
@@ -0,0 +1,136 @@
...
+            this.currentProduct.addProperty("warning", warnings);
+            this.currentProduct.addProperty("info", infos);
+
+            this.currentProduct.add("messages", this.currentMessages);
+
+            if (getWriter() != null) {
+                getWriter().println(gson.toJson(this.currentProduct));
+                getWriter().flush();
+            } else {
+                LOG.info(gson.toJson(this.currentProduct));
```
nutjob4life, line 76: I think if no writer is configured, `LOG.info` can add timestamps, log levels, class names, or other formatting cruft, which would make each line no longer NDJSON, right?

src/main/java/gov/nasa/pds/validate/report/StreamingJsonReport.java
```diff
@@ -0,0 +1,136 @@
...
+        msg.addProperty("message", problem.getMessage());
+
+        if (problem.getLineNumber() != -1) {
+            msg.addProperty("line", problem.getLineNumber());
+            if (problem.getColumnNumber() != -1) {
+                msg.addProperty("column", problem.getColumnNumber());
+            }
+        }
+        this.currentMessages.add(msg);
+    }
```
nutjob4life, line 47: Both `append` methods asusme `begin(Block.LABEL)` has already initialized `currentProduct` and `currentMessages`. If the report lifecycle ever emits problems outside a label block, this could throw a `NullPointerException`. May be worth guarding or making the lifecycle assumption explicit?

nutjob4life (changes requested):
Noticed a few things and left interspersed comments. What do you think? If these are all non-issues after all, by all means, let's move forward. Thanks in advance!

jordanpadams (commented):
## Code Review

Good approach overall. A few issues to address before merge:

---

**`StreamingJsonReport.java:122` — `summarizeProds` drops `skipped`**
The `skipped` parameter is never written to the summary JSON. `JSONReport` correctly emits it. Add `summary.addProperty("skipped", skipped)`.

---

**`ValidateLauncher.java:1055` — use `InvalidOptionException`, not `Exception`**
The PR touches this line (error string changed). The project already has `InvalidOptionException` imported and used in this same class (lines 344, 347, 357, 364) for exactly this case. Replace `throw new Exception(...)` with `throw new InvalidOptionException(...)`.

---

**`StreamingJsonReport.java:72,129` — `getWriter() != null` is always true**
`Report.writer` is initialized inline to `new PrintWriter(new OutputStreamWriter(System.out))` — it's never null. The `LOG.info(...)` fallback branches are dead code. Either remove the null check or document the intent.

---

**`StreamingJsonReport.java:3` — use SLF4J, not `java.util.logging`**
All other report classes use `org.slf4j.Logger` / `LoggerFactory`. Using JUL here bypasses the project's logging configuration and any MDC context.

---

**No way to distinguish product validation from referential integrity validation in the log stream**
When running on a bundle, two JSON records are emitted for the same LIDVID — one for product-level validation and one for referential integrity. Both look identical (same `label`, same `lidvid`, potentially both `PASS`). The full-text report separates these visually with section headers like "Product Level Validation Results" and "PDS4 Bundle Level Validation Results", but the streaming-json output has no equivalent. A consumer cannot tell which record is which.

Suggest adding a `"validationType"` field (e.g. `"product"` / `"referential"`) to each product JSON record so log consumers can filter or route accordingly.

---

**`--help` output does not list `streaming-json` as a valid option**
The `-s`/`--report-style` help text currently reads:

```
-s,--report-style <full|json|xml>   ...Valid values are 'full', 'json', and 'xml'...
```

This needs to be updated to include `streaming-json`, otherwise users have no way to discover the new flag.

---

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

jordanpadams (changes requested):
See comments above.

jordanpadams:
Moving this to draft since it needs updates and has been moved to a lower priority

